PR #33, opened 2015-11-10
Add license info in poms and THIRD-PARTY.properties

joshovi:
to run license info type
mvn license:aggregate-add-third-party


mcjaeger:
like it, but I have doubts, because our license is GPL 2.0 with class path exception not just GPL, which makes a substantial difference.

mcjaeger:
had closed this because the licensing issue needs to be revamped anyway. I am sorry zou have invested this work, but I think it makes more sense to start over with the colleagues.

Issue #16, opened 2015-11-02
Projects - Linked Projects


heydenreich:
From Mail Frederik Berck (Juli 23, 2015):

Show versions of linked projects.
Nice to have:  hyperlink to the linked project.

PR #50, opened 2015-11-27
Added project link version

joshovi:
Added project link version, and made project name clickable as redirekt to linkee


heydenreich:
closes #16

PR #51, opened 2015-11-30
Make type visible for component

### LINKED ISSUE #13: Component type adjustable for each release
From Mail Frederik Berck (Juli 23, 2015)

Make component type visible.
Use it for component filter.
Make component type (OSS, COTS...) adjustable for each release.
(This is also necessary in order to deal with dual licenses. Then the software can be created separately for each license model. )

heydenreich:
Add type to Component filter
Add Type to Component view and detail view
closes #13

### DIFF
--- a/frontend/sw360-portlets/src/main/java/com/siemens/sw360/portal/tags/OutTag.java
+++ b/frontend/sw360-portlets/src/main/java/com/siemens/sw360/portal/tags/OutTag.java
@@ -66,12 +66,12 @@ public class OutTag extends OutSupport {
         return super.doStartTag();
     }
 
-    //TODO remove uneeded...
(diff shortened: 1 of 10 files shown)


frontend/sw360-portlets/src/main/java/com/siemens/sw360/portal/tags/OutTag.java
```diff
@@ -66,12 +66,12 @@ public int doStartTag() throws JspException {
         return super.doStartTag();
     }
 
-    //TODO remove uneeded...
+    //TODO remove unneeded...
```
maximilianhuber: if not needed: remove code
if needet: remove comment

PR #53, opened 2015-11-30
Change the Versionstring to 1.1.0-SNAPSHOT

maximilianhuber:
* it was 0.1.1-SNAPSHOT
* the used command was `find . -type f -iname pom.xml -exec sed -i * 's/0.1.1-SNAPSHOT/1.1.0-SNAPSHOT/g' {} +`

**grep** says, that this were the only locations, where the versionstring was defined/used

### DIFF
--- a/backend/pom.xml
+++ b/backend/pom.xml
@@ -24,7 +24,7 @@
     <parent>
         <groupId>com.siemens.sw360</groupId>
         <artifactId>sw360</artifactId>
-        <version>0.1.1-SNAPSHOT</version>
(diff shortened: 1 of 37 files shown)


joshovi:
Hi,
maybe this is useful:
http://maven.apache.org/maven-release/maven-release-plugin/

Things that need to be done at a release:
1. Tag the commit
2. Create deliverables without SNAPSHOT in them (to do that typically you change all the pom.xml)
3. Bump up the snapshot version

PR #62, opened 2015-12-21
added approvers to attachments

### LINKED ISSUE #3: Attachments / Uploads should have approvers
When uploading attachments, an approver should be able to say "approve" to an upload item.

As approvers, the moderators would be suitable. The approve function is especially targeted for reports that are uploaded to the sw360portal, as an attachment to a release.

The implementation of the approve function involves a button, as the clicking user is presumed to be the approver. (no need to allow for text entry of an approver e-mail to enter other approvers) .

At every attachment item, the approver (e-mail) and the date of approval is shown. Please not use the time, as this bears problems with emloyment laws.

NOTE - this has an implication on the claring status counted for the projects: then the last number should not cound "report available", but "approved".

heydenreich:
fix delete attachments
closes #3


mcjaeger:
tested and tested in UI, works

PR #75, opened 2016-01-22
update lar files for singleTomcat with permissions

heydenreich:
remove old lar files


mcjaeger:
hm, for the import of the public pages, I find a lot of the application under choose application - is that correct?

screen shot 2016-01-22 at 14 56 32

mcjaeger:
apart from that: SW360 menue Admin is visible for admin but not users works!

Issue #60, opened 2015-12-10
Adding date picker for date fields


mcjaeger:
Apparently the argument that chrome has this build in does not count so much: Firefox users (and IE and Safari) have a hard time entering the date via keyboard.

A date picker for those fields would be super helpful. Please mind a compatible license for choosing the code, in the ideal case, it is part of liferay alloy!

Attached an image as example for dates in the project edit view.

![036 oss sw360 20151210 screenshot adding date widget](https://cloud.githubusercontent.com/assets/3418977/11720836/626a5ce4-9f61-11e5-8360-4acfa0cc2f47.png)

joshovi:
-----BEGIN PGP SIGNED MESSAGE-----
Hash: SHA1

Hi,

there was some promising discussion on

https://bugzilla.mozilla.org/show_bug.cgi?id=825294

about a month ago. Well, if it stays that way we can't help but
implement an own datepicker.

Best regards,
Johannes

On 12/10/2015 05:14 PM, Michael wrote:
> Apparently the argument that chrome has this build in does not
> count so much: Firefox users (and IE and Safari) have a hard time
> entering the date via keyboard.
> 
> A date picker for those fields would be super helpful. Please mind
> a compatible license for choosing the code, in the ideal case, it
> is part of liferay alloy!
> 
> Attached an image as example for dates in the project edit view.
> 
> ![036 oss sw360 20151210 screenshot adding date
> widget](https://cloud.githubusercontent.com/assets/3418977/11720836/626a5ce4-9f61-11e5-8360-4acfa0cc2f47.png)
>
>  --- Reply to this email directly or view it on GitHub: 
> https://github.com/siemens/sw360portal/issues/60
> 
-----BEGIN PGP SIGNATURE-----
Version: GnuPG v2.0.22 (GNU/Linux)

iQIcBAEBAgAGBQJWaaZYAAoJECKaE+bQwFiJNtoQAI4Oh37zjgwhexu6mtfrxBg0
2a1ZYyqv1Vef9Tu7PxlUEFbz3Hotr15OE2tD2ISbvabqfRr/Aa4ZhBn+RP5R4QTX
H6Mhvm+JFrA5abQnZtMyXSFap2HCnlNaICTUDQwUFFQXnOWai09SBCHrhdb501Nf
LtVk/j4J+nKtPm/Et5L98VMJRHo3w/ddOpAuMScl77aKuPHZl7LBmKY03F/+8QWB
[...]

PR #76, opened 2016-01-25
Dev/datepicker

sw360-blue-developer:
date picker (out of jquery) added for date fields in 'add project' mask, solves #60


maximilianhuber:
I think it would be good to squash these commits to a single one to have a clean history.

Issue #71, opened 2016-01-20
addTODOsBroken


heydenreich:
the add TODO feature in licenses portlet is probably broken

PR #78, opened 2016-02-01
Fix todos part1

heydenreich:


maximilianhuber:
This is a partial fix of #71

PR #79, opened 2016-02-01
feature(pom.xml): add plugins for tagging and setting versions

maximilianhuber:
This adds the plugins mentioned in https://github.com/siemens/sw360portal/wiki/Dev-releasing-SW360

They will wrap the process of releasing and tagging.


mcjaeger:
looks good go ahead with merge from my side

PR #85, opened 2016-02-08
If there are no problems theres no need to mention any

AngelM1981:
As a developer I want only read necessary documentation so that I can get fast grasp.


mcjaeger:
I think it is good to have some space already designated where such point could be named, if there is something in future. But thanks for you pull request.

PR #112, opened 2016-03-09
Attachments

heydenreich:


libraries/lib-datahandler/src/main/java/com/siemens/sw360/datahandler/couchdb/AttachmentConnector.java
```diff
@@ -78,4 +87,20 @@ public void deleteAttachments(Collection<Attachment> attachments) {
         connector.deleteIds(attachmentContentIds, AttachmentContent.class);
     }
 
+    public String getSha1FromAttachmentContentId(String attachmentContentId) {
+        try {
+            AttachmentContent attachmentContent = getAttachmentContent(attachmentContentId);
+            InputStream attachmentStream = getConcatenatedAttachmentPartsStream(attachmentContent, Integer.parseInt(attachmentContent.getPartsCount()));
+            byte[] attachmentBytes = null;
+            try {
+                attachmentBytes = IOUtils.toByteArray(attachmentStream);
```
fogninid: This will load the whole attachment in memory.
Something like [DigestUtils.html#sha1Hex(InputStream)](https://commons.apache.org/proper/commons-codec/apidocs/org/apache/commons/codec/digest/DigestUtils.html#sha1Hex%28java.io.InputStream%29) will calculate the sha1 per block

heydenreich: Good point. I'll change it.

PR #120, opened 2016-03-15
Fixed moderation request for adding attachments ...

### LINKED ISSUE #97: Moderation Requests for Attachments
There are problems with moderation requests for attachments:
1. When deleting an attachment, the corresponding moderation request cannot be displayed properly to the moderator and the Moderation Portlet is "temporarily not available"
2. When uploading an attachment the attachment is not added to the moderation request and even if the moderator accepts, the attachment is not in the database.

heydenreich:
... to projects, components and releases.

Also multiple add-actions and delete-actions can be added to the same moderation request.
LeadArchitect is removed from list of recipients of project moderation requests,
since he is only a contributor, not a moderator, and therefore is not allowed to delete
closes #97


maximilianhuber:
Code looks good

PR #135, opened 2016-04-11
refactored license shortnames to ids and removed mainLicenseNames fro…

### LINKED ISSUE #19: Refactor Main Licenses in Components and Releases
(original ticket from Johannes) 

Components and Releases have two fields:
* optional set<string> mainLicenseIds, 
* optional set<string> mainLicenseNames,

The Component fields are filled from Releases on Release change.

The names are never read, and set apart from the ComponentImporter.

The display of licenses is done via the DisplayLicense Tag, which looks up the whole license with the license client via the Id.

This is inconsistent and redundant and a solution would be to refactor licenses such that their Ids are the short names (so that the short names have to be unique, which they should anyway be). Then the two fields can be aggregated into one field.

The price would be that one can not easily change the shortname of a license after referencing it, but this rarely happens. Also it is easy to keep synchronous with fossology because there the license Id is the shortname.

alex-evo:
…m releases and components. Closes #19.

To migrate existing database data to the new schema, execute scripts/migrations/001_migrate_license_shortname_to_id.py (couchdb-python is a prerequisite)


scripts/migrations/001_migrate_license_shortname_to_id.py
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/python
```
maximilianhuber: Please add file header

scripts/migrations/001_migrate_license_shortname_to_id.py
```diff
@@ -0,0 +1,48 @@
...
+        del linking_obj['mainLicenseNames']
+    db.save(linking_obj)
+
+print 'Copying licenses to new ids'
+for license_row in licenses:
+    license = license_row.value
+    db.delete(license)
+    del license['_rev']
+    del license['shortname']
+    license['_id'] = license['shortname']
```
maximilianhuber: As far as I understand `del` you have deleted the value `license['shortname']` in the line above its usage

alex-evo: Yep. Must have flipped these two lines accidentally before commit.

PR #137, opened 2016-04-12
List Components with Main License

### LINKED ISSUE #20: List Components with Main License
Based on issue #19, a main use case asked for is to list components in the component main view by their main license (e.g. show only components with Apache-2.0 as main license)

At the same time, please limit the description to one line and truncate with three dots after a couple of chars (maybe 32?) and display full text with Mouse over.

At the same time, please change the default row view number from 10 to 25.

alex-evo:
Now listing components with their licenses in the main component view list and also allowing to search components specifically by their licenses.
Closes #20.


mcjaeger:
review-by: michael.c.jaeger@siemens.com

Testing not done, testing on clean system good. today I would like to see it with productive data set.

PR #133, opened 2016-03-23
Attachment handling

heydenreich:
Improved attachment handling:

 - component,project,release: uploading/deleting attachments without linking them instantaneously to the document
 - change on document attachments is only made when pressing "update document"
 - moderation request is only created when pressing "update document"
 - delete attachments from database when attachments are unlinked from document

- store uploaded attachment ids in a history per user and per documentId
- delete attachments from db on cancel and on delete document (also attachments that are
uploaded but not yet linked to document)
- properly handeled: user leaves edit mode by cancel, update or delete document
- not properly handled: user leaves edit mode by going to a different portal: in this case, uploaded attachments cannot be linked to document any more, but stay in db (use database sanitation to delete them later)
- closes #116

PR #139, opened 2016-04-14
Moderation restructure

### LINKED ISSUE #118: Write moderation request as diff to database
At the moment, a moderation request (MR) contains the entire document - the document as it is in the db at the moment the MR is created -with the changes applied that the user wants to make. 

When a moderator evaluates the moderation request, the document in the db might have been changed in the meantime. In this case, the difference between the db-document and the document from the moderation request is misleading:

- changes that the requesting user wanted to make disappear if the same changes have already been made to the db-document.
- fields or attachments that the user did not want to change may appear as changes if the fields or attachments have changed in the db-document. They are changed back if the moderation request is accepted.
-Especially, the moderator does not know whether the user wanted to delete the attachment or if it just has been added to the db-document after the moderation request was created.

Quick and dirty solution: moderation requests can only be accepted if the revision hash at the document in the MR equals the revision hash of the db-document.

Better solution: change the way MR's are written to the database. Instead of saving the entire document, save only the fields that have been changed with the new value. For attachments maintain two lists: the new attachments with all attachment meta data and a list of deleted attachments. The latter needs only to contain the attachmentContentId of the attachments to delete.

heydenreich:
Please merge only after #133 , since this builds upon that branch.
Closes #118 
	
Moderation request changed for projects, components, releases, licenses:
- documentAdditions and documentDeletions representing document changes are written to db:
- documentAdditions and documentDeletions contain: 
   - all fields that change according to mod. request
   - all required fields of the document 
   - all boolean and int fields of the document (are automatically initialized)
- edit view for document is displayed properly (including changes according to own moderation requests)
- accept moderation request works (diff is applied to document in database)
- merge.jsp displays difference between database document, deleted fields, added/changed fields
- for licenses, only todos and whitelists can be moderated (other changes may only be made by clearing admins and admins)


mcjaeger:

tested different variants of datatypes and in/between modifications (successfully)

Issue #127, opened 2016-03-17
ReleaseClearingStateSummary in project/detail


heydenreich:
1.  Release Clearing State Summary in project-details view misses the 5th field: number of approved releases.
2. If setting the clearing state of a release to "new" manually, then the release clearing state summary of the project is updated (in project view and project detail), but in the wrong way: instead of increasing the number of new releases, the number of releases "under clearing by project team" is increased.

Reason: the release holds also a map " ClearingTeamToFossologyStatus", where for each team the fossology status is listed. This map isn't updated when the clearing state is set manually.

Solution: the clearing state of a release should not be set manually, but should be automatically set.  (Already done for "report available" and "approved". Todo: for "under clearing"). 
The map  " ClearingTeamToFossologyStatus" can probably vanish if clearingRequests are introduced, see #128.

heydenreich:
1.  Release Clearing State Summary in project-details view misses the 5th field: number of approved releases.
2. If setting the clearing state of a release to "new" manually, then the release clearing state summary of the project is update, but in the wrong way: instead of increasing the number of new releases, the number of releases "under clearing by project team" is increased.

heydenreich:
part 1. is solved in PR#143

mcjaeger:
will provide a status diagram for this.

mcjaeger:
[NEW_CLEARING] -> [SENT_TO_FOSSOLOGY](user sends to fossology)

[SENT_TO_FOSSOLOGY] -> [UNDER_CLEARING](fossology status update queried by user)


[UNDER_CLEARING] -> [REPORT_AVAILABLE](user has created / uploaded attachment with type clearing report)

[REPORT_AVAILABLE] -> [APPROVED](attachment of type clearing report is approved using the controls in edit mode)

[APPROVED] -> [REPORT_AVAILABLE](approval for attachment of type clearing report is rejected
)
[APPROVED] -> [NEW_CLEARING](attachment of type clearing report is deleted)

[REPORT_AVAILABLE] -> [NEW_CLEARING](attachment of type clearing report is deleted)

alex-evo:
@mcjaeger, is there a general rule for dealing with multiple clearing reports?

mcjaeger:
Hello, seems like testing reveals that the issue is a bit more complicated. Let's keep it manually and then move it to 1.4 to have 1.3 out soon.

mcjaeger:
[...]

PR #143, opened 2016-04-15
display release clearing state summary in project/details view ...

heydenreich:
… with number of approved releases
solves part of #127 (part 1.), but not complete ticket


alex-evo:
tested ok.

PR #148, opened 2016-04-20
autocomplete for component categories

### LINKED ISSUE #82: Adding predefined categories to type (of component): DOAP
In some earlier  version there was a pull down menue, like as with programming languages for the component types.

So the user did not have free form, but a list to select from.

That should be back again. And is should based on an already established standard:

https://projects-old.apache.org/categories.html

(adding "other" mght be useful as well)

If could be implemented the same way it is done with platforms or programming languages.

alex-evo:
Added autocomplete suggestions for component categories.
The list of suggested categories can be edited in sw360.properties
Closes #82


mcjaeger:
the pull request appears rather small in terms of LOC what are the conflicts actually?

alex-evo:
the conflicts were in PortalConstants and sw360.properties. rebased/resolved.

mcjaeger:
review-by: michael.c.jaeger@siemens.com
actually also tested changing the list. Did direct edit of the two files because of having library and Library entries, but actually commit history looks bad - squash it?
other wise it could be merged directly.

alex-evo:
Github can squash when merging, can it not?

PR #179, opened 2016-06-15
fixed display and editing of a multiline project description

### LINKED ISSUE #152: New Line in Project Description Text
If there is a new line in the description text (just hit return key in the description field when editing it): The table view will fail - see screenshot below. In Firefox Web console, clicking on the line with the error will display the JSON array with the entry that caused the error.

It must be checked if the entered return is also leading to the same error (and fixed). Fix could be that for display in the table, all returns are cleaned out.

036 oss sw360 20160422 screenshot proejct with errornous description fail

alex-evo:
* fixed display and editing of a multiline project description
* moved DisplayDescription to the correct package folder
* fixed abbreviation in OutTag

closes #152


mcjaeger:

PR #183, opened 2016-06-17
add thrift service and pages for displaying vulnerabilities per release and per component

stefanjaeger-evo:


maximilianhuber:
Looks good to me up to:

  - some inline-`if` statements,
  - some `catch (Exception e)` blocks and 
  - we have already discussed about the verbose logging of the function [`assertEquals`](https://github.com/sw360/sw360portal/pull/183/files#diff-bcf931c76c5f085b518f45536b7184deR130)  which is used by @alex-evo to compare the entered passwords in the sing-up-portlet. 
  > If someone makes a mistake when entering his password or the retyped one, both will be written to the log!

  In fact might all changes in `SW360Assert.java` be unnecessary for this PR.

alex-evo:
Now that you mention it, that password comparison looks stupid to me. `String.equals()` would be just fine. Need to change that.

stefanjaeger-evo:
assertEquals is changed back

PR #187, opened 2016-06-24
allow property files to be placed at `/etc/sw360/`

bs-jokri:
the files placed at /etc/sw360/ overwrite the values in the other property files


mcjaeger:
review-by:michael.c.jaeger@siemens.com
(maybe macosx support would be nice, but it is not so important)

PR #195, opened 2016-07-13
LicenseInfo service

### LINKED ISSUE #180: Copyright and License Info Service for CLI/SPDX Attachment Parsing
In order to process CLI / SPDX2 documents, a service implement the following flow:

* take attachment id
* read (CLI / SPDX2) attachment, and ensure it is SPDX2 or CLI
* extract all copyright statements
* extract all license texts found there
* returning the two lists

maybe (?) the spdx tools are helpful for parsing SPDX2:
https://github.com/spdx/tools
http://spdx.org/tools/spdx/consolidated-spdx-tools-and-library

### LINKED ISSUE #181: Extensions for Copyright and License Info Service
Provide an interface call for release and then the attachment service fishes out the attachment type / info.

Priorities:
* Preferring SPDX over CLI
* Writing into result object actual file used
* if there are several files of the same type, resulting merged list will have distinct entries for copyrights and licenses only.

### LINKED ISSUE #182: Project Service can call Copyright and License Info service to aggregate CLIs
#Goal is to have an aggregated text file for the 

* copyrights
* licenses

of all involved releases, where involved is about the releases of a project and the releases of the linked subprojects.

Text file listing could be separated by release name (e.g. Apache Tomcat 7.0.54).

User can click a button in the linked project tree view to generate this text file.

alex-evo:
provides a licenseinfo service to collect copyrights and licenses information for a release and generate a licenseinfo file for a project. includes parsers for CLI and SPDX attachments.
closes #180 
closes #181 
closes #182


backend/src/src-cliservice/src/main/java/com/siemens/sw360/cliservice/parsers/CopyrightLicenseInfoParser.java
```diff
@@ -0,0 +1,29 @@
...
+public abstract class CopyrightLicenseInfoParser {
+    //    private static final Logger log = Logger.getLogger(CLIHandler.class);
+    public static final String FILETYPE_PARSING_IMPOSSIBLE = "NONE";
+    protected final AttachmentConnector attachmentConnector;
+
+    protected CopyrightLicenseInfoParser(AttachmentConnector attachmentConnector){
+        this.attachmentConnector = attachmentConnector;
+    }
+
+    public abstract boolean isApplicableTo(AttachmentContent attachmentContent);
```
maximilianhuber: Wouldn't it be better to have the full `Attachmant` instead of only `AttachmentContent` to decide whether a parser is applicable?

backend/src/src-cliservice/src/main/java/com/siemens/sw360/cliservice/CLIHandler.java
```diff
@@ -0,0 +1,121 @@
...
+    public CopyrightLicenseInfo getComponentLicenseInfoForAttachment(String attachmentContentId) throws TException {
+        AttachmentContent attachmentContent = attachmentDatabaseHandler.getAttachmentContent(attachmentContentId);
+        return applyParsers(attachmentContent);
+    }
+
+    @Override
+    public CopyrightLicenseInfo getComponentLicenseInfoForRelease(Release release) throws TException {
+        try {
+            CopyrightLicenseInfo result = nullToEmptySet(release.getAttachments()).stream()
+                    .filter(a -> a.getAttachmentType() == AttachmentType.COMPONENT_LICENSE_INFO_XML)
```
maximilianhuber: This filtering should be done by `isApplicableTo`

alex-evo: No. This is a draft.

Yes.

Yes, I think there was.

I had this idea of separating filtering which attachments are worth considering at all from actually checking whether they can be parsed. Now I see another problem with it. I'll rewrite this.

maximilianhuber:
I do not think that `cliservice` or `CLI` in general is a good name for something not related to the command line.

What about `readmeservice` or `disclosureDocumentService`?

backend/src/src-cliservice/src/main/java/com/siemens/sw360/cliservice/parsers/CopyrightLicenseInfoParser.java
```diff
@@ -0,0 +1,29 @@
...
+    public abstract CopyrightLicenseInfo getCLI(AttachmentContent attachmentContent);
```
maximilianhuber: What about returning `Optional<CopyrightLicenseInfo>` to signal parsing problems?

maximilianhuber: If this function would return `Optional<CopyrightLicenseInfo>` one could also drop `isApplicableTo` and just return `Optional.empty()` if a parser does not want to parse a file.

alex-evo: I'm going to create another thrift structure to transfer the status over the wire. `Optional<T>` would work.

There's one feature missing yet. The service should decide on one type of parser first and then parse only applicable files. Information from SPDX and CLI must not be mixed together. One can't drop `isApplicableTo yet`, otherwise one would possibly perform some unnecessary parsing.

maximilianhuber: I had not been aware of the restriction, that only one type of parser should be used.

mcjaeger: haha! naming discussion can be fun. :-)

Just a few thoughts on the purpose of this service:

* The ultimate solution for sw360 is to create a combined SPDX for the project or product and provide this with the customer. I am not sure you agree but at least that is my idea of it.
* Readme or disclosure document is just an artefact of the past, used for non-streamlined supply chains. The future is to have fully parse-able exchange formats. We use the readme now because we are so slow.
* Given the current service, why create another service for SPDX combined files then? You would use the same service and extend it, because half of the code is there anyways: look into the attachments of every involved release, parse the SPDX, create an aggregation based on the releases of a project.

Then with the naming:

* In the beginning there are service names, which were just substantives: Components, Projectsm Vendors, ...: nice and uniform.
* Actually, in the beginning we have had also the idea of a rest based API where the URL would result from such service, such as
GET http://localhost/components/mycomponent01 -> returns the JSON formatted dataset of mycomponents
GET http://localhost/components/ -> returns a listing of all component short names (IDs) etc
* PUT http:// ... can be done also nicely!
(* So in this world, how about bdpImport, why did not we call call projectsimport or import/projects, so you can mike a nice PUT request ...?)

Given these points, my thoughts on the naming for the current form:

* readme:  most people understand something else with readme than just license and copyright information. it is not the classic readme that introduces the software as first step. not so good.
* disclosure document: hm, maybe OK, but still too much to an artefact that seems to be organisation specific. Actually, I have never heard the term disclosure document before. hm.
* Component License Information: admitted a little bit bulky, we have hadused this term before. Maybe coincidental with another acronym. Frankly, I have seen so many projects with acronym overlap, I am not concerned much with the 427th occurrence of an acronym overlap. OK, maybe also not good.

I am voting for a name that reflects the future how about
* license information (easy compromise, just skip component)
* software data exchnage (too long)
* SPDX
* maybe if it is called licenseinformation it could be an idea to skip the word service as we have had also not -service suffix in the other services

maximilianhuber: The function `mergeCLIs` should then also be rewritten to only accept CLIs from the same parser and set `filetype` of the CLI

backend/src/src-cliservice/src/main/java/com/siemens/sw360/cliservice/CLIHandler.java
maximilianhuber: Yes, of course, I know that this is a draft. This was just some input while trying to think my way through your code.

alex-evo:
@mcjaeger, please have a look at the current state, if you have time. It already produces a file given some releases with CLI attachments.
Your feedback would be very much welcome.

mcjaeger:
did tests with generated CLI also with one level nesting: seems to work. Will further enhance nesting with multilevel nesting.

mcjaeger:
tested also with two levels of hierarchy ... works.
Suggestion: Change the file name to "LicenseInfo-[Projectname]-[YYYYMMDD].txt"

mcjaeger:
@maximilianhuber have noticed the following message: "... WARN  ListedLicenses:358 - Unable to access the SPDX listed licenses at http://www.spdx.org/licenses.  Using local file copy of SPDX listed licenses" -> which local list?

maximilianhuber:
Oh, I think SPDX is trying to get the current list of licenses and is not able to get them due to proxy problems.

In this case would the "local list" be a dump of the spdx licenses wich were added to the source repository of SPDX-Tools.

maximilianhuber:
See: https://github.com/spdx/tools/tree/master/resources/stdlicenses

alex-evo:
IMO, this PR is ready to be reviewed/tested/merged.

mcjaeger:
tested various combinations of hierarchzy and releases with no attachments and nonsense attachments -> all OK. @maximilianhuber as discussed, please provide more information about the SPDX file type handling

mcjaeger:
adding the following test:
* adding all six files from the SPDX tools samples directory to individual releases each
* adding all six releases to a project
* hitting "download license information"
* results in "(No license information available)" for all of the six releases
* tried with attachment types "document" and "component license information"

alex-evo:
I tested with one file (https://github.com/spdx/tools/blob/master/Examples/SPDXRdfExample-v2.0.rdf) and got the license information out of it. The only problem I had is that generating the file the first time after deployment took forever.

mcjaeger:
actually I feel like Alex explanation is too short: which document type? (a question I have asked earlier above, but still unanswered) @maximilianhuber could you pls retest or confirm it is working?

maximilianhuber:
Tested on fresh install of https://github.com/sw360/sw360portal/commit/5309bfc3bec45430f35f67d7264473cf1f7e7fef:

1. create component
2. create corresponding release
3. add some example spdx report from the six examples as **Document** attachment
4. create project
5. linked release to project
6. downloaded CLI via the button `Download License Information`

The downloaded text file contained license configuration, thus it seems to work.

--

The spdx-parser does not at the moment looks at the attachment type. But it looks at the attachment content type, which is set by the system. This might be inconsistent and depending on the host system / used browser. I will remove that since it is not necessary.

mcjaeger:
tested-by:michael.c.jaeger@siemens.com,maximilian.huber@tngtech.com
review-by:maximilian.huber@tngtech.com,Alex.Borodin@evosoft.com

PR #196, opened 2016-07-13
added directly linked releases of the project to linkedProjects

alex-evo:
Renamed tables and tabs to better reflect current state.


alex-evo:
@mcjaeger, there you go. It turned out to be easier than I first thought.

PR #200, opened 2016-07-19
make CPE ID editable

bs-jokri:
...and correct formula for CPE ID generation in the tooltip


alex-evo:
tested-by:alex.borodin@evosoft.com
reviewed-by:alex.borodin@evosoft.com

PR #201, opened 2016-07-19
removed wiki-pages for components and projects

bs-jokri:
removes the ui code related to the wiki entries. can be added as soon as we have a real implementation for that feature


alex-evo:
reviewed-by: alex.borodin@evosoft.com
tested-by: alex.borodin@evosoft.com

PR #202, opened 2016-07-19
licenses view.jsp adjusted to display licenses with ids with special …

bs-jokri:
…characters


alex-evo:
tested-by: alex.borodin@evosoft.com
reviewed-by: alex.borodin@evosoft.com

PR #208, opened 2016-08-19
Autosetting clearing state#127

### LINKED ISSUE #127: ReleaseClearingStateSummary in project/detail

alex-evo:
completed automatic management of release clearing state and removed the option of setting it manually.
closes #127

### DIFF
--- a/backend/src/src-fossology/src/main/java/com/siemens/sw360/fossology/handler/FossologyFileHandler.java
+++ b/backend/src/src-fossology/src/main/java/com/siemens/sw360/fossology/handler/FossologyFileHandler.java
@@ -28,10 +28,7 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
 import java.io.InputStream;
-import java.util.Collection;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Set;
+import java.util.*;
 
 import static com.google.common.base.Predicates.equalTo;
 import static com.google.common.base.Strings.isNullOrEmpty;
@@ -107,6 +104,7 @@ public class FossologyFileHandler {
                 //to avoid race conditions, get the object again
                 Release release = assertNotNull(componentService.getReleaseById(releaseId, user));
                 setFossologyStatus(release, clearingTeam, FossologyStatus.SENT, Integer.toString(fossologyUploadId), attachmentContent.getId());
+                updateReleaseClearingState(release, FossologyStatus.SENT);
                 updateRelease(release, user, componentService);
                 return RequestStatus.SUCCESS;
             }
@@ -144,6 +142,7 @@ public class FossologyFileHandler {
 
     private void updateFossologyStatus(Release release, User user, String clearingTeam, FossologyStatus currentStatus, ComponentService.Iface componentService) throws TException {
         setFossologyStatus(release, clearingTeam, currentStatus);
+        updateReleaseClearingState(release, currentStatus);
         updateRelease(release, user, componentService);
     }
 
@@ -180,6 +179,8 @@ public class FossologyFileHandler {
                     updateInDb = false;
                 }
             }
+            Optional<FossologyStatus> maxStatus = nullToEmptyMap(release.getClearingTeamToFossologyStatus()).values().stream().max(FossologyStatus::compareTo);
+            updateReleaseClearingState(release, maxStatus);
 
             if (updateInDb) {
                 updateRelease(release, user, componentClient);
@@ -189,6 +190,28 @@ public class FossologyFileHandler {
         getReleaseAndUnlockIt(releaseId, user, componentClient); // just unlockit
 
         return release;
+    }
+
+    private void updateReleaseClearingState(Release release, FossologyStatus fossologyStatus) {
+        updateReleaseClearingState(release, Optional.of(fossologyStatus));
+    }
+    private void updateReleaseClearingState(Release release, Optional<FossologyStatus> fossologyStatus) {
+        Optional<ClearingState> newClearingState = mapFossologyStatusToClearingState(fossologyStatus);
+        if (newClearingState.isPresent() && newClearingState.get().compareTo(release.getClearingState()) > 0){
+            release.setClearingState(newClearingState.get());
+        }
+    }
+
+    private Optional<ClearingState> mapFossologyStatusToClearingState(Optional<FossologyStatus> fossologyStatus) {
+        if (fossologyStatus.isPresent()){
+            if (fossologyStatus.get()==FossologyStatus.IN_PROGRESS){
+                return Optional.of(ClearingState.UNDER_CLEARING);
+            } else if (fossologyStatus.get().compareTo(FossologyStatus.SENT) >= 0 &&
+                    fossologyStatus.get().compareTo(FossologyStatus.IN_PROGRESS) < 0){
+                return Optional.of(ClearingState.SENT_TO_FOSSOLOGY);
+            }
+        }
+        return Optional.empty();
     }
 
     protected Release getReleaseAndUnlockIt(String releaseId, User user, ComponentService.Iface componentClient) throws TException {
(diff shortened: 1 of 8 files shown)


adrian-evo:
Tested and it is working as expected.

backend/src/src-fossology/src/main/java/com/siemens/sw360/fossology/handler/FossologyFileHandler.java
```diff
@@ -28,10 +28,7 @@
 import org.springframework.stereotype.Component;
 
 import java.io.InputStream;
-import java.util.Collection;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Set;
+import java.util.*;
```
mcjaeger: (do not change that, just a remark) actually I like identifying individual objects more since the loaded objects are less? - but after years I am not sure about it any more and one must consistently doing this to have this as an effect.

alex-evo: That's a readability optimization that IntelliJ did for me.
`import` is a compile-time instruction. It makes no difference on the compiled byte-code. In particular, it does not affect how many classes are loaded at runtime. You can have your code without imports at all, if you like to spell out fully qualified class names every time, or you can import all the packages with an asterisk import.

backend/src/src-fossology/src/main/java/com/siemens/sw360/fossology/handler/FossologyFileHandler.java
```diff
@@ -180,6 +179,8 @@ public Release getStatusInFossology(String releaseId, User user, String clearing
                     updateInDb = false;
                 }
             }
+            Optional<FossologyStatus> maxStatus = nullToEmptyMap(release.getClearingTeamToFossologyStatus()).values().stream().max(FossologyStatus::compareTo);
```
maximilianhuber: newlines would make this line of code more readeable, i.e.
```
            Optional<FossologyStatus> maxStatus = nullToEmptyMap(release.getClearingTeamToFossologyStatus())
                .values()
                .stream()
                .max(FossologyStatus::compareTo);
```

It might also be good to encapsulate this in a function.

backend/src/src-fossology/src/main/java/com/siemens/sw360/fossology/handler/FossologyFileHandler.java
```diff
@@ -191,6 +192,28 @@ public Release getStatusInFossology(String releaseId, User user, String clearing
         return release;
     }
 
+    private void updateReleaseClearingState(Release release, FossologyStatus fossologyStatus) {
+        updateReleaseClearingState(release, Optional.of(fossologyStatus));
+    }
+    private void updateReleaseClearingState(Release release, Optional<FossologyStatus> fossologyStatus) {
+        Optional<ClearingState> newClearingState = mapFossologyStatusToClearingState(fossologyStatus);
```
maximilianhuber: if you replace this with a call of the form
```
    Optional<ClearingState> newClearingState = fossologyStatus.flatMap(...)
```
the function `mapFossologyStatusToClearingState` would become
```
    private Optional<ClearingState> mapFossologyStatusToClearingState(FossologyStatus fossologyStatus) {
```
and you do not have to extract the content of the optional manually.

frontend/sw360-portlet/src/main/java/com/siemens/sw360/portal/portlets/components/ComponentPortletUtils.java
```diff
@@ -69,6 +65,9 @@ public static void updateReleaseFromRequest(PortletRequest request, Release rele
                         release.setReleaseIdToRelationship(new HashMap<String, ReleaseRelationship>());
                     updateLinkedReleaesFromRequest(request, release.releaseIdToRelationship);
                     break;
+                case CLEARING_STATE:
+                    // skip setting CLEARING_STATE. it is supposed to be set only programmatically, never from user input.
```
maximilianhuber: A short question:

 Doesn't this make it harder, for users which want to use SW360 without FOSSology?

@mcjaeger: what do you think?

libraries/lib-datahandler/src/main/java/com/siemens/sw360/datahandler/businessrules/ReleaseClearingStateSummaryComputer.java
```diff
@@ -50,7 +50,10 @@ public static ReleaseClearingStateSummary computeReleaseClearingStateSummary(Lis
     }
 
     static Void addReleaseWithStates(ReleaseClearingStateSummary summary, ViewedState globalState, ViewedState myTeamState, ViewedState otherTeamState) {
-        if (globalState.present()) { // if the release has its clearing state field set use it and ignore the rest
+        // if the release has its clearing state field set to REPORT_AVAILABLE or APPROVED use it and ignore the rest
+        // otherwise, look at what fossology has to say
+        // if all else fails, default to NEW
+        if (globalState.present() && globalState.compareTo(ViewedState.REPORT_AVAILABLE) >= 0) {
```
maximilianhuber: This is exactly what `Optional.filter` does:

>**Parameters:**
predicate - a predicate to apply to the value, if present
**Returns:**
an Optional describing the value of this Optional if a value is present and the value matches the given predicate, otherwise an empty Optional

this could make the code shorter.

frontend/sw360-portlet/src/main/java/com/siemens/sw360/portal/portlets/components/ComponentPortletUtils.java
maximilianhuber: Or is this just FOSSology related?

maximilianhuber:
Functionality is tested and works as expected.

backend/src/src-fossology/src/main/java/com/siemens/sw360/fossology/handler/FossologyFileHandler.java
alex-evo: Oh, yes! I hate that IDEA always reformats it to one line

backend/src/src-fossology/src/main/java/com/siemens/sw360/fossology/handler/FossologyFileHandler.java
alex-evo: thanks for the tip

frontend/sw360-portlet/src/main/java/com/siemens/sw360/portal/portlets/components/ComponentPortletUtils.java
mcjaeger: understand the point, but "sw360 is designed to work with fossology". Also, it could be OK, my points:
* clearing state is not really fossology dependent, but tool dependent.
* if there is a new tool, integration should be done accordingly (see also the ticket #128 )
* if there is no tool, no status shall be displayed there
* for the manual state, there is mainline state
does this sound acceptable?

libraries/lib-datahandler/src/main/java/com/siemens/sw360/datahandler/businessrules/ReleaseClearingStateSummaryComputer.java
alex-evo: Too bad `globalState` is not `Optional`

PR #213, opened 2016-08-30
fixed NullPointerExceptions in Moderation Requests

bs-jokri:
* added missing import in projects/merge.jsp
* Projects, releases, components imported from a datasource outside SW360 have several fields not set/null that are usually initialized when
generating the document in the SW360-UI. This caused NullPointerExceptions. This problem is fixed now.


mcjaeger:
an example how to test this?

bs-jokri:
@heydenreich please correct me if I am wrong. But this can be tested the following way:
1. Open Project import
2. type as server "mock"
3. import one example project
4. perform a change as another user that creates a moderation request
5. Try to do accept the moderation request as initial user

heydenreich:
Correct, this is the way to test it. (In the current version, already displaying the MR goes wrong in some cases. If you can view and accept the MR, everything is alright.)

PR #214, opened 2016-08-30
lics: import custom properties

bs-jokri:
This PR allows to import lics archives with licenses that have custom properties. For that a new table is introduced that is contained as csv file hi the archive.

### DIFF
--- a/frontend/sw360-portlet/src/main/java/com/siemens/sw360/portal/tags/DisplayMap.java
+++ b/frontend/sw360-portlet/src/main/java/com/siemens/sw360/portal/tags/DisplayMap.java
@@ -49,9 +49,13 @@ public class DisplayMap extends SimpleTagSupport {
 
     public static String getMapAsString(Map<String, String> map) {
         StringBuilder sb = new StringBuilder();
-        sb.append("<ul style=\"list-style-type: none; padding:0; margin:0;\">");
+        sb.append("<ul class=\"mapDisplayRootItem\">");
         map.entrySet().stream().forEach(e -> sb.append(
-                "<li><b>" + e.getKey() + "</b> " + e.getValue() + "</li>"
+                "<li><span class=\"mapDisplayChildItemLeft\">"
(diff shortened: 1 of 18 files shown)


heydenreich:
Must be rebased - wait for cveSearch PR.

heydenreich:
Rebased, result can be found on rebase/customProperties. bsinno-branch has to be updated to update this PR.

mcjaeger:
OK, maybe some more information is necessary actually to have this tested. Apart from "Further properties" as a column in the todo view, I am not seeing how the custom property map is actually used - since the detail.jspf seems to be the only location referencing it. Is it by import? (tested with Firefox by the way)

heydenreich:
Yes, you need a .csv-file that you can import. I'll send you one.

frontend/sw360-portlet/src/main/java/com/siemens/sw360/portal/tags/DisplayMap.java
```diff
@@ -49,9 +49,13 @@ public void doTag() throws JspException, IOException {
 
     public static String getMapAsString(Map<String, String> map) {
         StringBuilder sb = new StringBuilder();
-        sb.append("<ul style=\"list-style-type: none; padding:0; margin:0;\">");
```
mcjaeger: you could check if referencing class would be possible.

libraries/lib-datahandler/src/main/thrift/licenses.thrift
```diff
@@ -271,4 +273,8 @@ service LicenseService {
      * return filled todo
      **/
     Todo getTodoById( 1: string id);
+
+    list<CustomProperties> getCustomProperties(1: string documentType);
+
+    RequestStatus updateCustomProperties(1: CustomProperties customProperties, 2: User user )
```
mcjaeger: super nitpicking, pls consider whitespace before brackets close (because the other brackets are this way)

PR #215, opened 2016-08-30
add couchdb password / authentication support 

bs-jokri:
This PR improves the way the backend services and couch db communicate


maximilianhuber:
I will have a look on the compile error

maximilianhuber:
the rebase has solved the compilation problems

maximilianhuber:
@bs-jokri: on the repo sw360/sw360portal is the branch [dev/addCouchdbPW/contrib](https://github.com/sw360/sw360portal/tree/dev/addCouchdbPW/contrib) which is a rebased version of the bsinno brach. You could use it to reset this PR.

maximilianhuber:
@bs-jokri: the branch, mentioned in the last comment is updated and rebased on the current master. You now can use it to update your PR

frontend/sw360-portlet/src/main/java/com/siemens/sw360/portal/common/AttachmentPortletUtils.java
```diff
@@ -71,12 +71,7 @@ public AttachmentPortletUtils(ThriftClients thriftClients) {
...
             try {
-                AttachmentService.Iface client = thriftClients.makeAttachmentClient();
-                DatabaseAddress address = client.getDatabaseAddress();
-                connector = new AttachmentStreamConnector(address, downloadTimeout);
-            } catch (TException e) {
-                log.error("Could not get database address from attachment client...", e);
-                throw e;
+                connector = new AttachmentStreamConnector(downloadTimeout);
             } catch (MalformedURLException e) {
                 log.error("Invalid database address received...", e);
```
mcjaeger: it would be nice to have the DB url printed out here to trace back where an invalid setting might come from.

maximilianhuber: The information is no longer present at this point. To log this debug information one would have to catch the exception earlier or pass the information.

We will have a look

mcjaeger:
works good on vagrant-based setup!

Only thing is that password protecting the couchdb should take place AFTER the test were running on the integration setup. Otherwise the tests cannot create test databases, because they do not consider the password setting (which is not there in vagrant). But that is not a point for the merging of this PR.

mcjaeger:
docker works as well

PR #216, opened 2016-08-31
fix(#151): display hash with `<span>` instead of `<input>`

### LINKED ISSUE #151: Checksum field for attachments is too small
Somehow, the screen layout can be easily set into a mode where the checksum is truncatedly displayed. It would be good to have the complete checksum in most cases - for super small screens of course not, but checksum is truncated at fixed width, regardless of the screen width (for example, truncated at 1920 width) (tested with IE11)

maximilianhuber:
closes #151

### DIFF
--- a/frontend/sw360-portlet/src/main/webapp/css/sw360.css
+++ b/frontend/sw360-portlet/src/main/webapp/css/sw360.css
@@ -321,3 +321,8 @@ input[readonly].clickable {
 .release-list-entry-element input {
     width: 100%;
 }
+
+.hashvalue {
+    word-break: break-all;
+    font-family: monospace;
+}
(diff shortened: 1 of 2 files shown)


frontend/sw360-portlet/src/main/webapp/css/sw360.css
```diff
@@ -321,3 +321,8 @@ input[readonly].clickable {
 .release-list-entry-element input {
     width: 100%;
 }
+
+.hashvalue {
+    word-break: break-all;
+    font-family: monospace;
+}
```
heydenreich: I guess that the class "checksumInput" is not used any more after removing these lines. Maybe we should remove it in the css, too.

heydenreich:
Code looks good (apart from a single question).

PR #219, opened 2016-08-31
Set default project visibility to "groups and moderators"

### LINKED ISSUE #211: Change default project visibility
The default visibility for projects is now:
   `private`
and should be rather
   `group and moderators`
because the private setting is not useful in 90% of the cases

pfaffelmoser:
closes #211


frontend/sw360-portlet/src/main/java/com/siemens/sw360/portal/portlets/projects/ProjectPortlet.java
```diff
@@ -633,6 +638,10 @@ private void prepareProjectDuplicate(RenderRequest request) {
         return projects;
     }
 
+    private Visibility getDefaultVisibility() {
+        return DEFAULT_VISIBILITY;
+    }
+
```
heydenreich: Why not use the constant directly? It is within the same class.

heydenreich:
Tested, works fine.

PR #226, opened 2016-09-05
Dev/make bdpimport optional/backend/dev

bs-jokri:
removes the bdp functionality that is already moved to a separate repository


bs-jokri:
Contains wrong commits

PR #227, opened 2016-09-05
move bdpimport-backend to external package: remove it from sw360

bs-jokri:
Removes the bdp functionality from the backend that is already be moved to a separate repository.


heydenreich:
rebased - ready to test and merge

mcjaeger:
at import page, I noticed that there is still "BDP" in the pop up menu. It does not bother me much, but in case you intended to eliminate this I wanted to mention that.

maximilianhuber:
At the moment, a new backup service (i.e. bdp) has no option to tell the project import portlet in the frontend that it is there, i.e. register itself. This list thus consists of the only valid option, which is currently available.

Once additional import services emerge, we have to think about an registration mechanism.

mcjaeger:
I did also functional testing and just found the issue with the "BDP". Let's merge it, but who "all of you" is review-by: ?

alex-evo:
There isn't much to review, frankly - a bunch of files deleted. I reviewed it. The right side looks much cleaner.

PR #230, opened 2016-09-06
Added confirmation dialogues for deletion process

pfaffelmoser:
- Introduced jquery-confirm for confirm dialogue / alerting
- Added confirm dialogue for deletion of projects, components, releases, etc. via edit page
- Added hint on contained projects, releases and attachments to confirm message

Do not merge before dev/projectExcelExport and dev/datePicker_#171 !

closes #156

PR #233, opened 2016-09-14
increase lucene filter limit and avoid moderation request zombie

### LINKED ISSUE #114: moderation request zombie
if the last moderator of a moderation request clicks "remove me from moderators" the moderator will be removed. in this case the moderation request still exists in the database and cannot be accepted or declined by another moderator.

solutions can be:
* the button "remove me from moderator" will be disabled (including tooltip) or hidden
* the moderation request will be declined with or without notice

### LINKED ISSUE #221: Project list: filter use results in lists limited to 25 entries
(requires testing because seen in commit version 32780d36). If the project filter is applied with a group value (formerly department), then the resulting listing is limited to 25 entries (although there are many more entries).

heydenreich:
set lucene.limit to 150  …
* get a maximum of 150 results when using the filter in project and component portlet
* has to be tested: is speed still ok?
* closes #221 if speed is ok

avoid moderation request zombie  …
* if last moderator wants to unsubscribe, pop-up - not allowed
* closes #114

* merge only after #230 and use ..../jquery-confirm.min.css and ../jquery-confirm.min.js from #230

### DIFF
--- a/frontend/sw360-portlet/src/main/java/com/siemens/sw360/portal/portlets/moderation/ModerationPortlet.java
+++ b/frontend/sw360-portlet/src/main/java/com/siemens/sw360/portal/portlets/moderation/ModerationPortlet.java
@@ -72,10 +72,10 @@ public class ModerationPortlet extends FossologyAwarePortlet {
         RemoveModeratorRequestStatus status = null;
         try {
             status = client.removeUserFromAssignees(id, user);
+            request.setAttribute(PortalConstants.REQUEST_STATUS, status);
         } catch(TException e) {
             log.error("Error in Moderation ", e);
         }
-        request.setAttribute(PortalConstants.REQUEST_STATUS, status);
         renderRemoveModerationRequestStatus(request, response, status);
     }
 
--- a/frontend/sw360-portlet/src/main/java/com/siemens/sw360/portal/portlets/moderation/ModerationPortlet.java
+++ b/frontend/sw360-portlet/src/main/java/com/siemens/sw360/portal/portlets/moderation/ModerationPortlet.java
@@ -14,6 +14,7 @@ import com.siemens.sw360.datahandler.common.CommonUtils;
 import com.siemens.sw360.datahandler.common.SW360Constants;
 import com.siemens.sw360.datahandler.common.SW360Utils;
 import com.siemens.sw360.datahandler.thrift.ModerationState;
+import com.siemens.sw360.datahandler.thrift.RemoveModeratorRequestStatus;
 import com.siemens.sw360.datahandler.thrift.attachments.Attachment;
 import com.siemens.sw360.datahandler.thrift.components.Component;
 import com.siemens.sw360.datahandler.thrift.components.ComponentService;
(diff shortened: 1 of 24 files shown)


maximilianhuber:
@heydenreich: please rename the title

heydenreich:
- rebased and changed in order to re-use (entire) project detail view also in moderation view 
- before: linkedProjects.jspf was already used in project details and moderation request view 
- crashed in moderation request view because of missing licenseDownloadInfo
- this problem was solved, licenseDownloadForm is not displayed any more in moderation view

mcjaeger:
tested the moderator thing with components, projects and releases -> OK

mcjaeger:
tested the search limit -> works as well

mcjaeger:
looked at the source code with my basic understanding, looks OK.

frontend/sw360-portlet/src/main/java/com/siemens/sw360/portal/portlets/moderation/ModerationPortlet.java
```diff
@@ -57,11 +58,27 @@
...
+        final User user = UserCacheHolder.getUserFromRequest(request);
+        final String id = request.getParameter(MODERATION_ID);
+        ModerationService.Iface client = thriftClients.makeModerationClient();
+        RemoveModeratorRequestStatus status = null;
+        try {
+            status = client.removeUserFromAssignees(id, user);
+        } catch(TException e) {
+            log.error("Error in Moderation ", e);
+        }
+        request.setAttribute(PortalConstants.REQUEST_STATUS, status);
```
maximilianhuber: Would be good to move this into the `try`-block. One should not set an attribute to `null`

libraries/lib-datahandler/src/main/java/com/siemens/sw360/datahandler/common/Moderator.java
```diff
@@ -25,6 +25,7 @@
 
 import static com.siemens.sw360.datahandler.common.CommonUtils.addAll;
 import static com.siemens.sw360.datahandler.common.CommonUtils.removeAll;
+import static com.siemens.sw360.datahandler.common.CommonUtils.nullToEmptySet;
```
maximilianhuber: this might be unwanted

frontend/sw360-portlet/src/main/webapp/html/moderation/releases/delete.jsp
```diff
@@ -52,13 +52,8 @@
...
 </p>
-<input type="button" onclick="acceptDelete()" id="edit" value="Accept Delete"    class="acceptButton">&nbsp;
-<input type="button" onclick="removeFromModerators()" id="edit" value="Remove Me from Moderators"    class="ignoreButton">&nbsp;
-<input type="button" onclick="declineDelete()" id="edit" value="Decline Deletion"    class="addButton">&nbsp;
-<input type="button" onclick="postPone()" id="edit" value="Postpone Decision"    class="postponeButton">&nbsp;
-<input type="button" onclick="cancel()" id="edit" value="Cancel"    class="cancelButton">
-
-
+<%@include file="/html/moderation/includes/moderationActionButtons.jspf"%>
+<br/>
```
maximilianhuber: This return looks like a hack to solve a css problem. Might be better to wrap the buttons into a div and use the `display:...` or `margin:...` properties

maximilianhuber: See https://github.com/sw360/sw360portal/pull/233/files#diff-972eef148286086787d19cde5b1b0038R335

frontend/sw360-portlet/src/main/java/com/siemens/sw360/portal/portlets/moderation/ModerationPortlet.java
heydenreich: done

libraries/lib-datahandler/src/main/java/com/siemens/sw360/datahandler/common/Moderator.java
heydenreich: done

frontend/sw360-portlet/src/main/webapp/html/moderation/releases/delete.jsp
heydenreich: it is already wrapped in a div. I removed the <br/>

PR #234, opened 2016-09-15
fix: SPDX licenseinfo should also use `Document` AttachmentType

maximilianhuber:
The `AttachmentType.Document` was previously excluded

In the effort to solve this bug, also some code was refactored.


heydenreich:
reviewed: OK

PR #235, opened 2016-09-15
Dev/simplify dependencies

koa:
moved all dependencies and the most plugins from intermediate poms to leaf poms (packaging != "pom")


mcjaeger:
hm, the branch has conflicts now, but nevertheless found it works with vagrant deployment. resulting deployment is working with component, release, attachment, project creation and generating readme.

mcjaeger:
@koa could you pls. help us with explaining why it is better to move plugins downwards (from backend/pom.xml to specific files)?

backend/src/src-bdpimport/src/test/java/com/bosch/osmi/sw360/bdp/datasink/thrift/ThriftUploaderTest.java
```diff
@@ -140,7 +141,7 @@ public void testIfIsAlreadyInDbWrapperEmpty() {
 
     @Test
     public void testIfIsAlreadyInDbWrapperEmptyList() {
-        Optional<String> toCheck = thriftUploader.searchExistingEntityId(Optional.of(Collections.EMPTY_LIST),
```
mcjaeger: (from discussion) so this is about best practice to prefer method over constant access? (fine, just for clarification)

koa: The Collections.emptyList() hast the element type T (new Style, since Java 1.5), the Collections.EMPTY_LIST has no element type (old Style since Java 1.2).
The Eclipse compiler cannot detect the correct type of Collections.EMPTY_LIST (and eclipse always recommend the style).
At runtime i expect no differences.

backend/svc/svc-fossology/pom.xml
```diff
@@ -40,13 +40,84 @@
...
+            <version>2.5</version>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>log4j</groupId>
+            <artifactId>log4j</artifactId>
+            <version>${log4j.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>junit</groupId>
```
maximilianhuber: At least the dependencies to `junit`, `hamcrest`, `mockito` ... are unused here (this part contains no tests).

This might also be true for other dependencies moved to the leaf poms.

koa:
I prefer to handle the modules as inependent as possible.
Every module declares exactly that what this module needs, not what the most of similar modules needs.
Aditionally its hard to analyze what exactly happens while compiling when there are so much pom files which have complicated override rules (im not sure in which order parents and pom imports ovverride each other).

If it is important to define the version of a plugin only once i prefer to use dependencyManagement.

Additionally i think there are too much modules and levels in this project. Normally multi-module projects has 1 level with 3-8 modules.I have never worked on a multi-module project with more than one deployment artifact.
Removing functionality from intermediate layers makes it easier to remove and/or refactor it.

mcjaeger:
I think the issue with the independent modules is understood.

We are just uncertain, because the pull request makes a flat copy of the super-dependecies to individual module pom.xml without checking if such dependency is actually needed at this level.

Granted it does not make it worse in terms of what build dependencies are, it is just maybe more confusing rather than having a bunch of dependencies at the top level of the backend project.

mcjaeger:
hello, while we were delighted by this contribution, it has some technical problems. I am closing this pull request because in its current state, there too much adaptations necessary before merging it. The main issue is that dedicated subproject pom.xml have the disadvantage that they hold all dependencies which is confusing, because in some subprojects only less than half of the dependencies are actually required by that subproject. And this is more confusing than its current state.  I intended to work on this but I do not find the time now, sorry. 

Please reopen the pull request if you wish, right now I would like to pull this off the list of open pull requests.

PR #243, opened 2016-09-26
Liferay error messages red

### LINKED ISSUE #49: Session message on error should be red (instead of green)
If one shows a banner using `com.siemens.sw360.portal.portlets.Sw360Portlet.setSessionMessage`,
one will see a green banner. I think that the color should depend on `requestStatus`, i.e.
* **green** only for `SUCCESS`
* **red** for `FAILURE`
* **yellow** for everything else 

#### screenshot
bildschirmfoto-vendors - sw360 - chromium
Banner is green but the message says "Vendor short **could not** be....".

heydenreich:
Liferay error messages:
* error message in SW360Portlet red instead of green
* Show error message instead of "portlet not available" for entering projects/components/release details cq. edit mode
* closes #49

Repair commits:
* set scheduling interval for cve-search back to 24 hours
* display projects/vulnerabilities.jsp again

Suggestions to test error messages:
* signup portlet: try to sign up with an email address (full name, external id) that is already in the system
* try to delete license that is used by a release
* enter project/component/release detail view for p/c/r that is meanwhile deleted
* enter project/component/release edit view for p/c/r that is meanwhile deleted


frontend/sw360-portlet/src/main/webapp/html/components/detail.jsp
```diff
@@ -43,41 +26,60 @@
     <portlet:param name="<%=PortalConstants.COMPONENT_ID%>" value="${component.id}"/>
 </portlet:resourceURL>
 
-<div id="header"></div>
-<p class="pageHeader"><span class="pageHeaderBigSpan">Component: ${component.name}</span>
-     <span class="pull-right">
+<c:catch var="attributeNotFoundException">
```
alex-evo: good idea, this catch

frontend/sw360-portlet/src/main/java/com/siemens/sw360/portal/common/ErrorMessages.java
```diff
@@ -0,0 +1,70 @@
...
+public class ErrorMessages {
+
+    public static final String PROJECT_NOT_ADDED = "Project could not be added.";
+    public static final String COMPONENT_NOT_ADDED = "Component could not be added.";
+    public static final String RELEASE_NOT_ADDED = "Release could not be added.";
+    public static final String ERROR_GETTING_PROJECT = "Error fetching project from backend.";
+    public static final String ERROR_GETTING_COMPONENT = "Error fetching component from backend.";
+    public static final String ERROR_GETTING_RELEASE = "Error fetching release from backend.";
+    public static final String LICENSE_USED_BY_RELEASE =  "Request could not be processed, as license is used by at least one release!";
+    public static final String DOCUMENT_USED_BY_PROJECT_OR_RELEASE = "Document could not be processed, as it is used by other Projects or Releases!";
```
heydenreich: This message is used in SW360Portlet, so in principle for various situations. I guess, you are right for now, but maybe be will use it for a situation where we cannot update or rename or so, because the document is used.

I don't know, what's better.

alex-evo: I do not think renaming or updating can be a problem in the future as we reference documents by id. This message is only used when an attempt to delete a document returns status `IN_USE`.
I mean, this is a minor issue, obviously. I just thought one can be more precise in the wording.

mcjaeger:
works for the 4 cases with me.

PR #244, opened 2016-09-26
display version of project in projects view

### LINKED ISSUE #220: Adding product version in the overview listing
In the main listing in the project view, the main listing is missing the project version (at most other occasions, the version is added to the project name. What happens is that users add project entries with the same name but not with the version number. As such, there will be a number of entries with the same name and users need to open the entry to find the actual one.

Attached a screenshot.

image002

Then there is a small layout issue which is that the version is added without a space character to the project name at the detail view (looks odd).

image001

heydenreich:
* closes #220


heydenreich:
closed as this PR is covered in PR #245

PR #246, opened 2016-09-26
ensured consistent closing of attachment InputStream

### LINKED ISSUE #217: Downloading attachments unreliable due to connection leak
Problem discovered while testing license info file generation with SPDX attachments. License info file could not be generated because the `SPDXParser` stumbled on a "ConnectionPoolTimeoutException: Timeout waiting for connection"

Restarting the server remedies the problem for a while.
Top suspect is the `AttachmentConnector.getSha1FromAttachmentContentId(...)` method, where the attachmentStream is not closed (wrapping with `AutoCloseInputStream` should do the trick), but it may have other accomplices.

alex-evo:
closes #217


mcjaeger:
works with 10 documents ... nice.

PR #248, opened 2016-09-28
Display using projects and components as table

### LINKED ISSUE #150: Used Projects in Components not easy to read for many projects
For a long list of projects in the component view, the listing of projects should be row-wise, because for a longer listing with popular components, the listing of projects in a "single line" is not easy to read.

At least carriage return should be possible. As possible enhancements, the department and the project responsible could be named.

heydenreich:
In project-details, project-edit, component-detail, release-detail, moderation request for projects (delete and change):
* all using projects are displayed with "name (version)",
also those not accessible for user
* all accessible projects have additionally
responsible and group displayed
* using components are displayed with vendor, main licenses and type
* closes #150


alex-evo:
@heydenreich, if you could rebase this PR, I'd be grateful

backend/src-common/src/main/java/com/siemens/sw360/components/summary/ProjectSummary.java
```diff
@@ -54,4 +57,32 @@ protected static void setSummaryFields(Project document, Project copy) {
             }
         }
     }
+
+    protected static void setDefaultFields(Project document, Project copy) {
+        copyField(document, copy, _Fields.ID);
```
alex-evo: As ID, NAME, and VERSION are always set and probably must always be set, I'd put them into an extra method like `setMinimalFiields`

backend/src-common/src/main/java/com/siemens/sw360/datahandler/db/ProjectRepository.java
```diff
@@ -168,7 +168,17 @@ public ProjectRepository(DatabaseConnector db) {
...
     public Set<Project> searchByLinkingProjectId(String id, User user) {
         Set<String> searchIds = queryForIdsByPrefix("bylinkingprojectid", id);
-        return new HashSet<>((makeSummaryFromFullDocs(SummaryType.SHORT, filterAccessibleProjectsByIds(user, searchIds))));
+        return linkedProjectSummaryAccordingToAccessibility(searchIds, user);
+    }
+
+    private Set<Project> linkedProjectSummaryAccordingToAccessibility(Set<String> ids, User user){
+        Set<Project> accessibleLinkedProjects = filterAccessibleProjectsByIds(user, ids);
+        Set<String>  accessibleLinkedProjectIds = accessibleLinkedProjects.stream().map(Project::getId).collect(Collectors.toSet());
+        Set<String> unaccessibleLinkedProjectIds = ids.stream().filter(pid -> ! accessibleLinkedProjectIds.contains(pid)).collect(Collectors.toSet());
```
alex-evo: There's `Sets.difference()` for this

frontend/sw360-portlet/src/main/webapp/html/projects/edit.jsp
```diff
@@ -55,6 +55,16 @@
         <jsp:include page="/html/utils/includes/attachmentsUpload.jsp"/>
         <jsp:include page="/html/utils/includes/attachmentsDelete.jsp" />
     </core_rt:if>
+    </p>
+    <core_rt:if test="${not addMode}" >
+        <input type="button" id="formSubmit" value="Update Project" class="addButton">
+        <input type="button" value="Cancel" onclick="cancel()" class="cancelButton">
```
alex-evo: Cancel button is added in both cases. Could be outside the `core_rt:if`s

backend/src-common/src/main/java/com/siemens/sw360/components/summary/ProjectSummary.java
heydenreich: Only the name is a required field according to the thrift struct. ID, NAME and VERSION are set in the "unaccessible..." and "accessible..." summary, because they have to be displayed (or used) in the linked project view. In principle that could change again and is independent of other summaries. And in the SUMMARY, I set all fields anyway by iterating over all fields - there would be no advantage in setting some minimal fields already before. And in all other cases (Summaries SHORT, EXPORT, DETAILED_EXPORT, HOME - no idea if they're all used, probably not), also the description and clearing team are set at the moment. So, I would prefer to leave it as it is and not to split the default case in a minimal case and some other fields needed for the default. - OK?

backend/src-common/src/main/java/com/siemens/sw360/datahandler/db/ProjectRepository.java
heydenreich: OK, I'll use that.

heydenreich: Although, that introduces guava again ...

frontend/sw360-portlet/src/main/webapp/html/projects/edit.jsp
heydenreich: True.

backend/src-common/src/main/java/com/siemens/sw360/components/summary/ProjectSummary.java
alex-evo: ok

backend/src-common/src/main/java/com/siemens/sw360/datahandler/db/ProjectRepository.java
alex-evo: There's also `Set.removeAll()` in case you don't like guava

PR #251, opened 2016-09-29
Improved links to projects and releases

### LINKED ISSUE #210: Improving links of projects to other projects or to releases
_Updates from discussion in [...]_

(In edit mode) When linking releases (and sub projects) to a project, the same release or sub project (or both) can be linked multiple times. The user should not be allowed to do that and a warning message should appear explaining that this request is not completed. [resolved, see comments below]

another effect is that if there is the same relation object (target) with a different project relation in two relations, the linked project view is not containing "the second relation" (whatever second relation means). Point here is to check if that is actually wanted/unwanted. [resolved, see comments below]

With this ticket a small modification could be applied: The link names "referred" and "contained" are not self explaining and should be renamed. No super proposals for rename but maybe: [remaining item]
* "is-sub-project" (contained)
* "is-related-project" (for referred)

Also relation unknown should not be default (but contained or is-sub-project). [remaining item]

alex-evo:
- Changed display strings of project and release relationships CONTAINED and REFERRED to more self-explaining 'Is a subproject' (resp. 'Is a subrelease') and 'Related'.
- Changed default relationship from UNKNOWN to REFERRED

closes #210

### DIFF
--- a/frontend/sw360-portlet/src/main/java/com/siemens/sw360/portal/portlets/components/ComponentPortlet.java
+++ b/frontend/sw360-portlet/src/main/java/com/siemens/sw360/portal/portlets/components/ComponentPortlet.java
@@ -290,7 +290,7 @@ public class ComponentPortlet extends FossologyAwarePortlet {
 
                 final String fullname = vendor != null ? vendor.getFullname() : "";
                 ReleaseLink linkedRelease = new ReleaseLink(release.getId(), fullname, release.getName(), release.getVersion());
-                linkedRelease.setReleaseRelationship(ReleaseRelationship.REFERRED);
+                linkedRelease.setReleaseRelationship(ReleaseRelationship.CONTAINED);
                 linkedReleases.add(linkedRelease);
             }
         } catch (TException e) {
--- a/frontend/sw360-portlet/src/main/java/com/siemens/sw360/portal/portlets/components/ComponentPortlet.java
+++ b/frontend/sw360-portlet/src/main/java/com/siemens/sw360/portal/portlets/components/ComponentPortlet.java
@@ -66,6 +66,7 @@ import static com.siemens.sw360.portal.common.PortletUtils.addToMatchedByHistogr
  * @author cedric.bodet@tngtech.com
  * @author Johannes.Najjar@tngtech.com
  * @author stefan.jaeger@evosoft.com
+ * @author alex.borodin@evosoft.com
  */
 public class ComponentPortlet extends FossologyAwarePortlet {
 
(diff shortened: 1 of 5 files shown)


frontend/sw360-portlet/src/main/java/com/siemens/sw360/portal/portlets/components/ComponentPortlet.java
```diff
@@ -289,6 +290,7 @@ private void serveNewTableRowLinkedRelease(ResourceRequest request, ResourceResp
 
                 final String fullname = vendor != null ? vendor.getFullname() : "";
                 ReleaseLink linkedRelease = new ReleaseLink(release.getId(), fullname, release.getName(), release.getVersion());
+                linkedRelease.setReleaseRelationship(ReleaseRelationship.REFERRED);
```
heydenreich: The ticket says that is-sub-project (=CONTAINED) should be the standard. For linked releases of releases, the ticket does not say anything. So just to check: is this really Ok here?

frontend/sw360-portlet/src/main/java/com/siemens/sw360/portal/portlets/projects/ProjectPortlet.java
```diff
@@ -287,7 +287,7 @@ private void serveNewTableRowLinkedProjects(ResourceRequest request, ResourceRes
             for (String linkedId : linkedIds) {
                 Project project = client.getProjectById(linkedId, user);
                 ProjectLink linkedProject = new ProjectLink(linkedId, project.getName());
-                linkedProject.setRelation(ProjectRelationship.UNKNOWN);
+                linkedProject.setRelation(ProjectRelationship.REFERRED);
```
heydenreich: But for projects the ticket says something: CONTAINED should be the default. Or do we have to change the ticket?

alex-evo:
I guess interracial relationships defy being shoehorned into the narrow box of our imaginations and therefore need expanded free-text descriptions ;-)
Seriously though, I have no idea why this is so.

frontend/sw360-portlet/src/main/java/com/siemens/sw360/portal/portlets/projects/ProjectPortlet.java
alex-evo: now changed to `CONTAINED`

frontend/sw360-portlet/src/main/java/com/siemens/sw360/portal/portlets/components/ComponentPortlet.java
alex-evo: changed to `CONTAINED`

PR #252, opened 2016-09-29
rename labels in release detail view

### LINKED ISSUE #189: Rename labels in release detail view
Rename 1:  "Licenses" in the attributes of releases sounds misleading, because there should be many licenses, as found in the clearing report. However, this is not helpful here, because the information that people looks for is what the general component  (release, I know) main license is. rename -> "Main Licenses"
Rename 2: "Release Date" -> "Release Date of this Release"
Rename 3: "Contacts" is like very questionable, because the user cannot see what is it for. Could be renamed in "Contact person in org" or better provide a mouse over "i" information. (similar to existing "i" mouseovers)
Rename 4: Provide also in "i" mouse over info why to set the mainline status or what the values mean (OPEN: new, MAINLINE: recommended use, SPECIFIC: tolerated use, PHASE OUT: avoid use)

Additionally in Release Clearing details (both edit and normal view):
* process start and evaluation date -> rename to "evaluation start" and "Evaluation end"
* and change the order of the fields (field start before/left field for finish)
* put request id and request add info together in a separate box "Request Info": (Request Id, Evaluation end, Additional Request, evaluation start )
* put all ECC fields together see #191 under topic "ECC Info": (AL, ECCN, Material Index Number, Assessor Contact Person, Assessor Department, ECC Comment)

### LINKED ISSUE #191: Rearrange Clearing Detail view from Release
The clearing detail view from release could be more structured:
1) actually clearing details only
2) ECC info
3) other
right know it contains everything and the ECC assessor person is assesor person only and the persons mix this with clearing "assessment".
Another example for this is the evaluation date, which is referring to the clearing information or the comment which is presumably from ECC info.

Also: Remove the checkbox settings for edit view in the clearing view, because that is actually depending if the attachments are there. So the check boxes can be shown in clearing details, but what is checked, depends actually on which attachments are present.

alex-evo:
- renamed fields in release detail view
- rearranged fields of release clearing details in display and edit mode

closes #189 
closes #191

### DIFF
--- a/frontend/sw360-portlet/src/main/webapp/html/components/includes/releases/clearingDetails.jspf
+++ b/frontend/sw360-portlet/src/main/webapp/html/components/includes/releases/clearingDetails.jspf
@@ -1,5 +1,5 @@
 <%--
-  ~ Copyright Siemens AG, 2013-2015. Part of the SW360 Portal Project.
+  ~ Copyright Siemens AG, 2013-2016. Part of the SW360 Portal Project.
   ~
   ~ All rights reserved. This program and the accompanying materials
   ~ are made available under the terms of the Eclipse Public License v1.0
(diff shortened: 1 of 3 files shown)


alex-evo:
@mcjaeger, please check if I got the arrangement of fields right. I tried to take into account descriptions in both #189 and #191.

mcjaeger:
it is all there (in display and edit view to close #189 and #191, but the mouseover for mainline state seems to be missing?

mcjaeger:
I will retest with mouse over to check if the tool tip is really missing

PR #258, opened 2016-10-04
add and cancel button in projects-edit were displayed twice - fixed

heydenreich:
* probably that was a rebase problem of the previous PR


mcjaeger:
please check with priority as this is a fix for broken master.

PR #259, opened 2016-10-04
ClearingAdmins and Admins added as moderators in moderation requests

### LINKED ISSUE #240: Add clearing_admin and admin for moderation requests
For the moderation requests, they are routed to the moderators and the creator. However, it should be also that

* for projects, clearing admin with same group membership should receive a moderation request as well. also the admin of the sw360 should receive the moderation request
* for components and releases, clearing admin and admin should receive moderation requests as well.

heydenreich:
* for projects:
 * ClearingAdmins of department of project
 * all admins

* for components and releases:
 * ClearingAdmins of department of creator
 * all admins

* closes #240


backend/src/src-moderation/src/main/java/com/siemens/sw360/moderation/db/ModerationDatabaseHandler.java
```diff
@@ -260,6 +272,12 @@ public void createRequest(User user) {
         addOrUpdate(request);
     }
 
+    private String getDepartmentFromUserEmail(String userEmail) throws TException {
```
alex-evo: I have a feeling that in this method you're doing the job of the `UserService` for it. Admittedly, it's only a two-liner job, so feel free to disagree.
Also, I'd call this method `getDepartmentByUserEmail`

backend/src/src-moderation/src/main/java/com/siemens/sw360/moderation/db/ModerationDatabaseHandler.java
```diff
@@ -280,24 +298,36 @@ public void createRequest(User user) {
...
+        List<User> resultingUsers = relevantUsersOfDepartment.isEmpty() ? allRelevantUsers : relevantUsersOfDepartment;
+        Set<String> resultingUserEmails = resultingUsers.stream()
                     .map(User::getEmail)
                     .collect(Collectors.toSet());
 
-        return adminEmails;
+        return resultingUserEmails;
+    }
+
+    private Set<String> getUsersAtLeast(UserGroup userGroup) {
```
heydenreich: Ok, I let one method call the other. But I have to treat the null case separately, since at least the setup-admin has department _null_.

backend/src/src-moderation/src/main/java/com/siemens/sw360/moderation/db/ModerationDatabaseHandler.java
```diff
@@ -280,24 +298,31 @@ public void createRequest(User user) {
...
-        List<User> allAdministrators = sw360users
+        List<User> allRelevantUsers = sw360users
                     .stream()
-                    .filter(user1 -> PermissionUtils.isUserAtLeast(UserGroup.ADMIN, user1))
+                    .filter(user1 -> PermissionUtils.isUserAtLeast(userGroup, user1))
                     .collect(Collectors.toList());
-        List<User> administrators = allAdministrators.stream()
+
+        List<User> relevantUsersOfDepartment = Collections.emptyList();
+        if(department != null) {
```
alex-evo: This does not make sense to me. If `department` is `null`, then `user.getDepartment().equals(department)` will return `false`. It's when `user.getDepartment()` is `null` that you get a problem. But that was already in the old code and seemed to work fine with the setup user

backend/src/src-users/src/main/java/com/siemens/sw360/users/UserHandler.java
```diff
@@ -88,4 +88,7 @@ public RequestStatus sendMailForAcceptedModerationRequest(String userEmail) thro
         return db.sendMailForAcceptedModerationRequest(userEmail);
     }
 
+    public String getDepartmentByEmail(String email) throws TException{
+        return getByEmail(email).getDepartment();
```
alex-evo: looks like a `NullPointerException` in the making

backend/src/src-moderation/src/main/java/com/siemens/sw360/moderation/db/ModerationDatabaseHandler.java
heydenreich: True. But we can leave it like this to avoid the work of filtering, when we already know that we do not want to filter.

backend/src/src-users/src/main/java/com/siemens/sw360/users/UserHandler.java
heydenreich: True. I'll try to avoid it.

backend/src/src-moderation/src/main/java/com/siemens/sw360/moderation/db/ModerationDatabaseHandler.java
alex-evo: sure

PR #262, opened 2016-10-06
scheduleAdminPortlet display improved

bs-jokri:
This PR adds the functionality missed in the previous PR about cve search functionality.

* buttons enabled/disabled depending on whether or not cvesearch is scheduled
* offset, interval and nextScheduledSync displayed as settings

autostart scheduling

* after deployment, the scheduling service schedules all services that are registered in
the autostart property in the sw360.properties file
* if property file or one of the properties (offset, interval) is not present use default value
* if one of the properties is invalid, log this and do not schedule service
* default: cvesearch service is _not_ auto-scheduled

### DIFF
--- a/backend/src/src-schedule/src/main/java/com/bosch/osmi/sw360/schedule/service/ScheduleHandler.java
+++ b/backend/src/src-schedule/src/main/java/com/bosch/osmi/sw360/schedule/service/ScheduleHandler.java
@@ -9,10 +9,12 @@
  */
 package com.bosch.osmi.sw360.schedule.service;
 
+import com.bosch.osmi.sw360.schedule.timer.ScheduleConstants;
 import com.bosch.osmi.sw360.schedule.timer.Scheduler;
 import com.siemens.sw360.datahandler.common.SW360Utils;
 import com.siemens.sw360.datahandler.permissions.PermissionUtils;
 import com.siemens.sw360.datahandler.thrift.RequestStatus;
+import com.siemens.sw360.datahandler.thrift.RequestStatusWithBoolean;
 import com.siemens.sw360.datahandler.thrift.RequestSummary;
 import com.siemens.sw360.datahandler.thrift.ThriftClients;
 import com.siemens.sw360.datahandler.thrift.schedule.ScheduleService;
@@ -20,6 +22,8 @@ import com.siemens.sw360.datahandler.thrift.users.User;
 import org.apache.log4j.Logger;
 import org.apache.thrift.TException;
 
+import java.util.Date;
+import java.util.Optional;
 import java.util.function.Supplier;
 
 public class ScheduleHandler implements ScheduleService.Iface {
@@ -42,7 +46,7 @@ public class ScheduleHandler implements ScheduleService.Iface {
             try {
                 return body.get();
             } catch (TException e) {
-                log.error("was not able to schedule sync for client with name:" + serviceName + " message:" + e.getMessage(), e);
+                log.error("Was not able to schedule sync for client with name:" + serviceName + " message:" + e.getMessage(), e);
                 return RequestStatus.FAILURE;
             }
         };
@@ -50,8 +54,9 @@ public class ScheduleHandler implements ScheduleService.Iface {
     }
 
     @Override
-    public RequestSummary scheduleService(String serviceName, User user) throws TException {
-        if (!PermissionUtils.isAdmin(user)){
+    public RequestSummary scheduleService(String serviceName) throws TException {
+        if(ScheduleConstants.invalidConfiguredServices.contains(serviceName)){
+            log.info("Could not schedule " + serviceName + " because of invalid configuration.");
             return new RequestSummary(RequestStatus.FAILURE);
         }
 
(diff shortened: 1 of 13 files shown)


heydenreich:
Maybe a hint for testing: 

* in the sw360.properties-file in src-schedule, set schedule.cvesearch.interval.seconds = 120 
(=every 2 minutes)
* in the same properties-file, uncomment the line autostart - cvesearchService
* after deployment, check in the admin/schedule-portlet that the schedule-button is inactive (as scheduling has been autostarted already) and the cancel-buttons are active
* via the bdp-mock, import the svm-test-project
* check that updating vulnerabilies has started after 2 minutes
* check that vulnerabilities were imported for the svm-test-project

mcjaeger:
so I can test?

heydenreich:
I got a "So I can test" - mail about this PR, but I do not see the comment here right now. Someone has to test, I guess - gave some hints. Of course, we have also tested it, but it's our PR from bsinno...

libraries/lib-datahandler/src/main/thrift/sw360.thrift
```diff
@@ -49,10 +49,10 @@ enum VerificationState {
...
 
 struct VerificationStateInfo {
-    1: required string checkedOn,
-    2: required string checkedBy,
-    3: optional string comment,
-    4: required VerificationState verificationState,
+  1: required string checkedOn,
+  2: required string checkedBy,
+  3: optional string comment,
+  4: required VerificationState verificationState,
```
alex-evo: These changes in spacing can be safely rolled back.

mcjaeger:
tested it first hand with straight build and deploy, looks good. however, changing the settings as described in the sw360.properties in the webapp/scheduler/WEB-INF/classes does not show the desired effect (I was so excited about what time offset will be displayed for 100000 seconds ...) any tricks? I will try again with a central /etc/sw360/sw360.properties file.

heydenreich:
Did you restart the scheduling service (or the server at all) after changing the settings?

backend/src/src-schedule/src/main/java/com/bosch/osmi/sw360/schedule/service/ScheduleHandler.java
```diff
@@ -42,16 +46,17 @@ private boolean wrapSupplierException(SupplierThrowingTException body, String se
...
                 return RequestStatus.FAILURE;
             }
         };
         return Scheduler.scheduleNextSync(wrappedBody, serviceName);
     }
 
     @Override
-    public RequestSummary scheduleService(String serviceName, User user) throws TException {
-        if (!PermissionUtils.isAdmin(user)){
+    public RequestSummary scheduleService(String serviceName) throws TException {
```
heydenreich: Because the SchedulingServlet has to call ` handler.scheduleService(cleanServiceName);` automatically in the autostart-method on startup.

backend/src/src-schedule/src/main/java/com/bosch/osmi/sw360/schedule/service/ScheduleHandler.java
```diff
@@ -89,4 +95,46 @@ public RequestStatus unscheduleAllServices(User user) throws TException {
...
+                .setAnswerPositive(answer);
+    }
+
+    private RequestStatusWithBoolean failedRequestStatusWithBoolean(){
+        return new RequestStatusWithBoolean().setRequestStatus(RequestStatus.FAILURE);
+    }
+
+    @Override
+    public int getFirstRunOffset(String serviceName){
+        return ScheduleConstants.SYNC_FIRST_RUN_OFFSET_SEC.get(serviceName) != null ? ScheduleConstants.SYNC_FIRST_RUN_OFFSET_SEC.get(serviceName) : -1;
```
heydenreich: I changed it.

mcjaeger:
I restarted the entire server. But I think I did not successfully edit the file. I am sorry. and the 10000 seconds translates to 27h 46 min 40secs. 
The buttons change accordingly.

libraries/lib-datahandler/src/main/thrift/sw360.thrift
alex-evo: I dunno about nicer, but you didn't add to consistent formatting, obviously. But ok, fine. Let's have it nicer at least in some places

mcjaeger:
But just a question: After pressing "Cancel Scheduled Cve Updates" or "Cancel All Scheduled tasks": I noticed the value for "next synchronization" does not jump back to zero or something like: nothing scheduled ...? 
I understand it like this: "Next synchronization" is like the time when it would kick off when the task is scheduled. I think the label is then maybe confusing, but I do not have a better proposal. So, but no big deal, let's do this later or as commit in another pull req.

heydenreich:
No, the display is merely about the configured settings. ("What would I get if I pressed the schedule-button now").

heydenreich:
Oh, now it's merged already - I was still busy refactoring according to Alex' comments...

heydenreich:
It was not about too big changes, though.

PR #264, opened 2016-10-07
fix broken project hierarchy and release hierarchy

alex-evo:
Both for projects and for releases, computing the relationship tree didn't allow referencing the same project/release twice anywhere in the relationship tree even when that didn't introduce cycles.

### DIFF
--- a/backend/src-common/src/main/java/com/siemens/sw360/datahandler/db/ComponentDatabaseHandler.java
+++ b/backend/src-common/src/main/java/com/siemens/sw360/datahandler/db/ComponentDatabaseHandler.java
@@ -618,17 +618,26 @@ public class ComponentDatabaseHandler {
         return ReleaseClearingStateSummaryComputer.computeReleaseClearingStateSummary(releases, clearingTeam);
     }
 
-    public List<ReleaseLink> getLinkedReleases(Map<String, ?> relations) {
+    public List<ReleaseLink> getLinkedReleases(Map<String, ?> relations, Map<String, Release> releaseMap) {
         List<ReleaseLink> out;
-
-        final List<Release> releases = releaseRepository.getAll();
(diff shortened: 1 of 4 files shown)


backend/src-common/src/main/java/com/siemens/sw360/datahandler/db/ComponentDatabaseHandler.java
```diff
@@ -646,10 +646,12 @@ public ReleaseClearingStateSummary getReleaseClearingStateSummary(Set<String> id
         return out;
     }
 
-    private Optional<ReleaseLink> createReleaseLink(Set<String> visitedIds, Map.Entry<String, ?> entry, String id, Release release, Map<String, Release> releaseMap, String parentId) {
-        if (visitedIds.add(id)) {
+    private Optional<ReleaseLink> createReleaseLink(Stack<String> visitedIds, Map.Entry<String, ?> entry, String id, Release release, Map<String, Release> releaseMap, String parentId) {
+        ReleaseLink releaseLink = null;
```
heydenreich: A lot of arguments ;-). If createReleaseLink is always called with id==entry.getKey() it's maybe worth considering to remove id from the argument list..

alex-evo: Prompted by your comment, I reviewed argument lists in all four related methods and cleaned up/reorganized them a little. See
[60f788e](https://github.com/sw360/sw360portal/pull/264/commits/60f788efda19d5bbb3c8456e901e91193a7713ee)

mcjaeger:
so we can merge? (after checkes have completed)

heydenreich:
I haven't tested.

PR #272, opened 2016-10-19
fix: linked projects and linked releases tree-tables expand wrong node

alex-evo:
fix: linked projects and linked releases tree-tables expand wrong node when the same project or release is present elsewhere in the table.

#### Background
For tree-table expanding/collapsing to work correctly, values of `data-tt-id` attribute must be unique within the table. Previously, every project or release could only occur once in the table and couchdb ids were used for `data-tt-id`, but PR #264 allowed the same object to be included multiple times in the table, which broke the expanding/collapsing functionality as `data-tt-id`s were not unique anymore.


mcjaeger:
I was tempted to merge this, but maybe someone from the BDP Import corner has a look, because the thrift data model has changed.

heydenreich:
It's ok. As long as the bdp-import is still part of SW360, you will realize if you break something. The two fields that you removed should be no problem for the bdp-import.

PR #285, opened 2016-11-23
Fixed Problems With Datatables Handling

### LINKED ISSUE #232: release deletion: ajax fails to remove row from data table
In the details view of a component, tab `release overview`:

if clicking on the trash bin to delete a release, the release is successfully deleted in the database, but the row is not removed in the UI.

When using dataTable  no error is written to the JavaScript-log. When using DataTable the error "row is not a function" is written to the log.
In all other view.jsps DataTable is used the same way and it works fine.

### LINKED ISSUE #268: View not refreshed automatically when deleting a release
When deleting a release from Release Overview list, the message "Do you really want..." appears, and when clicking OKAY, the deleted release is still listed among other releases (i.e. view not refreshed). When deleting a component, attachment, etc., the views are refreshed automatically.

### LINKED ISSUE #271: After deleting a project from any page, the view returns to first page
Under Projects tab navigate to any page different from First, and delete a project. Expected result is that the listed page is recreated and remains the same (if it is not the last and didn't contained single item). Actual result is that the first page is displayed instead which seems quite annoying, e.g. if the user wanted to delete or edit another project from the same page.

The same happens when there is a filter applied with multiple pages result.

alex-evo:
renamed all oTable JS vars to specific names to fix #268 and prevent similar potential problems. closes #268, closes #232

fixed resetting to the first page after rows are deleted from datatables. closes #271

### DIFF
--- a/frontend/sw360-portlet/src/main/webapp/html/admin/bulkReleaseEdit/view.jsp
+++ b/frontend/sw360-portlet/src/main/webapp/html/admin/bulkReleaseEdit/view.jsp
@@ -128,11 +128,11 @@
         load();
     });
 
-    var oTable;
+    var componentsInfoTable;
 
     //This can not be document ready function as liferay definitions need to be loaded first
     function load() {
-        oTable = configureComponentBasicInfoTable();
+        componentsInfoTable = configureComponentBasicInfoTable();
     }
 
     function configureComponentBasicInfoTable(){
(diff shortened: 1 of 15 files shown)


adrian-evo:
Works as expected for both fixed issues.

mcjaeger:
review-by:maximilian.huber@tngtech.com
tested-by:adrian.vaidos@siemens.com

mcjaeger:
review-by:maximilian.huber@tngtech.com
tested-by:adrian.vaidos@siemens.com

PR #290, opened 2016-11-29
Dev/ini contrib related p rs

bs-jokri:
Removed unused images, icons and text files


mcjaeger:
running ./sw360-install.sh seems to fail?

> ==> default: - [shell-provisioning] This is the SW360 installation script
> ==> default: - [shell-provisioning] Copy fossology keys to /sw360portal/backend/src/src-fossology/src/main/resources
> ==> default: cp: cannot create regular file ‘/sw360portal/backend/src/src-fossology/src/main/resources’: No such file or directory
> The SSH command responded with a non-zero exit status. Vagrant
> assumes that this means the command failed. The output for this command
> should be in the log above. Please read the output to determine what
> went wrong.

anyone else still doing the sw360vagrant?

PR #291, opened 2016-12-01
Contrib/rename packages

bs-jokri:
Package renamed to org.eclipse.sw360


mcjaeger:

PR #299, opened 2016-12-08
cveSearch: only allow matches that have consumed something

bs-jokri:
The levenstein distance metric which is used to check if a cve is applicable to a component release, is improved in this PR. It takes now a version mismatch into account, which was not considered before.

### DIFF
--- a/backend/src/src-cvesearch/src/main/java/org/eclipse/sw360/cvesearch/datasource/heuristics/SearchLevels.java
+++ b/backend/src/src-cvesearch/src/main/java/org/eclipse/sw360/cvesearch/datasource/heuristics/SearchLevels.java
@@ -168,7 +168,7 @@ public class SearchLevels {
 
         return vendorProductList.stream()
                 .map(match -> new NeedleWithMeta(cpeBuilder.apply(match.getNeedle()),
-                        "heuristic (dist. " + match.getDistance() + ")"))
+                        "heuristic (dist. " + (useVersionInformation ? "0" : "1") + match.getDistance() + ")"))
                 .collect(Collectors.toList());
     }
 
--- a/backend/src/src-cvesearch/src/main/java/org/eclipse/sw360/cvesearch/datasource/matcher/ModifiedLevenshteinDistance.java
+++ b/backend/src/src-cvesearch/src/main/java/org/eclipse/sw360/cvesearch/datasource/matcher/ModifiedLevenshteinDistance.java
@@ -9,8 +9,6 @@
  */
 package org.eclipse.sw360.cvesearch.datasource.matcher;
 
-import java.util.Optional;
-
 import static org.eclipse.sw360.datahandler.common.CommonUtils.nullToEmptyString;
 
 public class ModifiedLevenshteinDistance {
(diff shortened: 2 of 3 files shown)


backend/src/src-cvesearch/src/test/java/org/eclipse/sw360/cvesearch/datasource/matcher/ModifiedLevenshteinDistanceTest.java
```diff
@@ -12,6 +12,8 @@
 import org.junit.Test;
 
 import static org.eclipse.sw360.cvesearch.datasource.matcher.ModifiedLevenshteinDistance.levenshteinMatch;
+import static org.hamcrest.Matchers.*;
+import static org.junit.Assert.*;
 
 public class ModifiedLevenshteinDistanceTest {
```
alex-evo: I see that the behaviour of `levenshteinMatch` has changed so that the needle must match the haystack in at least one character, or else `Integer.MAX_VALUE` is returned and also the characters' case is ignored now, but neither of these changes is reflected in javadoc comments or in the tests. This Levenshtein distance trickery is not the most self-explanatory code in the project. I wish you'd update the documentation on what exactly is modified from the standard Levenshtein distance.

backend/src/src-cvesearch/src/main/java/org/eclipse/sw360/cvesearch/datasource/matcher/ModifiedLevenshteinDistance.java
```diff
@@ -9,13 +9,61 @@
...
     public static Match levenshteinMatch(String needle, String haystack){
         return new Match(needle,
-                calculateModifiedLevenshteinDistance(needle,nullToEmptyString(haystack).toLowerCase().replace(' ', '_')));
+                calculateModifiedLevenshteinDistance(needle, nullToEmptyString(haystack).replace(' ', '_')));
+    }
+
+    public static class LevenshteinCost {
+
+        private final int d;
+        private final boolean consumed;
```
alex-evo: Maybe, just maybe, this field and corresponding methods would better be called `matched`

backend/src/src-cvesearch/src/test/java/org/eclipse/sw360/cvesearch/datasource/matcher/ModifiedLevenshteinDistanceTest.java
```diff
@@ -22,7 +24,7 @@ public void testBasicRules() {
         String needDle = "needDle";
 
         // equal strings have distance 0
-        assert(levenshteinMatch(needle,needle).getDistance() == 0);
+        assertThat(levenshteinMatch(needle,needle).getDistance(), is(0));
```
alex-evo: This is nice. Why stop at only this assert?

mcjaeger:
Any support how to test this? maybe a joint conf call?

bs-jokri:
@mcjaeger the test would be that a release without CPE has an empty name and the distance is not matching to a very short component name in a CPE e.g. X11. But maybe we can just have a short telco and show the effect of this fix.

backend/src/src-cvesearch/src/main/java/org/eclipse/sw360/cvesearch/datasource/heuristics/SearchLevels.java
```diff
@@ -168,7 +168,7 @@ public SearchLevels addGuessingSearchLevels(CveSearchApi cveSearchApi, int vendo
 
         return vendorProductList.stream()
                 .map(match -> new NeedleWithMeta(cpeBuilder.apply(match.getNeedle()),
-                        "heuristic (dist. " + match.getDistance() + ")"))
+                        "heuristic (dist. " + (useVersionInformation ? "0" : "1") + match.getDistance() + ")"))
```
maximilianhuber: Yes, i want to make it a two-digit number starting with a `1` or a `0`, where `0` means that the version has matched.

backend/src/src-cvesearch/src/main/java/org/eclipse/sw360/cvesearch/datasource/matcher/ModifiedLevenshteinDistance.java
```diff
@@ -50,48 +98,49 @@ public static int calculateModifiedLevenshteinDistance(String needle, String hay
...
-        int[] curcost = new int[needleLength];
+        LevenshteinCost[] oldcost = new LevenshteinCost[needleLength];
+        LevenshteinCost[] curcost = new LevenshteinCost[needleLength];
 
-        for (int i = 0; i < needleLength; i++) oldcost[i] = i;
+        for (int i = 0; i < needleLength; i++) oldcost[i] = new LevenshteinCost(i);
 
-        int savedCostsWhenSkippedSpaceSeparatedPrefix     = 0;
-        int minimalCostsWhenSkippedSpaceSeperatedPostfix = Integer.MAX_VALUE;
+        int savedCostsWhenSkippedSpaceSeparatedPrefix                 = 0;
```
maximilianhuber: Sorry, i missed that. I will try to update the PR

PR #306, opened 2016-12-20
Download Source Code Bundle for a Project

bs-jokri:
This PR contains a feature that allows to download a source code bundle for a project similar to the generation of the notice file.


frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/common/AttachmentPortletUtils.java
```diff
@@ -84,23 +86,72 @@ private AttachmentStreamConnector getConnector() throws TException {
         return connector;
     }
 
+    protected InputStream getStreamToServeAnFile(List<AttachmentContent> attachments) throws TException, IOException {
```
alex-evo: naming: Serve**An**File - wrong article.

mcjaeger:
tested:
* hierarchy of projects
* multiple attachments
* large attachments
* various type attachments
* big attachments
* small attachments
* sweet attachments
* ugly attachments
...
nice that single file stays single but multiple files are in folder. also UI did not show any inconsistencies.

PR #307, opened 2016-12-20
Small fix in folder structure of tests in src-common

bs-jokri:
move tests `.../src/main/test/com/...` to `.../src/test/java/com...`


alex-evo:
I wonder why do we have two classes named `ProjectDatabaseHandlerTest` in the codebase? Can they be merged?

maximilianhuber:
You might be right, one should compare the classes. The files moved here were previously not even recognized by the compile process. Maybe they are already unnecessary.

PR #309, opened 2016-12-20
Some smaller contributions collected in one PR

bs-jokri:
- Replace findbugs.annotations in ektorp lib with Apache Licensed reimplementation
- Allow FOSSology keys to be placed at `/etc/sw360` 
   This also introduces a new function CommonUtils.loadResource which generalizes the placement of resources at /etc/sw360
- if `Name` of an SPDX license in imported spdx file is not filled, use its `LicenseId` as license name
- setting project default values for new projects
   - state = active
   - visibility = group+moderators
   - projectType = customer
   - moderators = []
   - contributors = []
   - clearing team = unknown
- set Liferay SessionError instead of SessionMessage for license duplicate
- generalize location of thrift binary
   since thrift gets installed into $PATH, there is no need to add the fully qualified path to the binary.

   Even the Ubuntu package installs it into another folder /usr/bin/ instead of /usr/local/bin/ and there are other location where the binary can be found.

### DIFF
--- a/libraries/lib-datahandler/src/main/java/org/eclipse/sw360/datahandler/common/CommonUtils.java
+++ b/libraries/lib-datahandler/src/main/java/org/eclipse/sw360/datahandler/common/CommonUtils.java
@@ -1,5 +1,7 @@
 /*
- * Copyright Siemens AG, 2014-2016. Part of the SW360 Portal Project.
+ * Copyright Siemens AG, 2014-2016.
+ * With modifications by Bosch Software Innovations GmbH, 2016
+ * Part of the SW360 Portal Project.
  *
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
@@ -23,6 +25,7 @@ import org.eclipse.sw360.datahandler.thrift.users.UserService;
 import org.apache.commons.csv.CSVFormat;
 import org.apache.commons.csv.QuoteMode;
 import org.apache.commons.io.FilenameUtils;
+import org.apache.commons.io.IOUtils;
 import org.apache.log4j.Logger;
 import org.apache.thrift.TException;
 import org.ektorp.DocumentOperationResult;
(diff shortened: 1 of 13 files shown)


mcjaeger:
I think we have conflicting default values here: for us, it is product and visibility public. How about having the default type being possible to set in sw360*.properties?

bs-jokri:
Good idea. @maximilianhuber what do you think?

maximilianhuber:
Yes that is a good idea. And we have to rebase this PR.

libraries/lib-datahandler/src/main/java/org/eclipse/sw360/datahandler/common/CommonUtils.java
```diff
@@ -505,6 +508,35 @@ public static Properties loadProperties(Class<?> clazz, String propertiesFilePat
         return props;
     }
 
+    public static Optional<byte[]> loadResource(Class<?> clazz, String resourceFilePath) {
+        return loadResource(clazz, resourceFilePath, true);
+    }
+
+    public static Optional<byte[]> loadResource(Class<?> clazz, String resourceFilePath, boolean useSystemRecourses) {
```
alex-evo: Like this.
btw, `useSystemRecourses` is spelled incorrectly

maximilianhuber:
@mcjaeger The way this PR currently defines the default visibility of projects does not allow configuration via a properties file. It is [hardcoded in the thrift file](https://github.com/sw360/sw360portal/pull/309/commits/7fe183dc9760b32a00387a219e0162c23614ceb7#diff-58a35ca111021fb800adea75699e6dcaR78).

mcjaeger:
I think I messed something up. It seems like it is Ok for now.

maximilianhuber:
There are new test failures. I will have a look and correct the rebase-failure.

mcjaeger:
@bs-jokri @maximilianhuber could you pls more helpful with some comments in terms of how to test this? so, "use SPDX-LicenseId as license name" is about the SPDXParser.java ?

mcjaeger:
@bs-jokri @maximilianhuber in case, could you add an spdx example file, it seems like the SPDXrdfExample.20 from the SPDX Java tools is not showing this effect properly - maybe, but I am not sure, what to actually look for. For example, for this file I am missing LGPL, but I am not sure, it is correct, if it would be listed?

maximilianhuber:
@mcjaeger I have updated the line related to your question. I hope that it is clearer now.

maximilianhuber:
@mcjaeger the appearance of LGPL is not affected by this PR. I think you have discovered an independent bug which has to be investigated.

mcjaeger:
how about adding test file. So I observed
* only spdx:licenseId -> this is used
* both spdx:licenseId and spdx:name -> name is used
* only spdx:name -> white page returned. and nullptr exception on server. @maximilianhuber have you observed this too?

PR #310, opened 2016-12-20
display non-existing user with email

bs-jokri:
- if user is stored in a field of a document, display user's email in details and edit view
- this way, in edit mode, the user is not deleted silently

### DIFF
--- a/frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/tags/DisplayUserEdit.java
+++ b/frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/tags/DisplayUserEdit.java
@@ -95,11 +95,13 @@ public class DisplayUserEdit extends NameSpaceAwareTag {
                     if (!Strings.isNullOrEmpty(email))
                         user = client.getByEmail(email);
                 } catch (TException e) {
-                    user = null;
+                    throw new JspException("Exception occurred in User client.");
                 }
+                emailList.add(email);
                 if (user != null) {
-                    emailList.add(email);
                     userList.add(user.getFullname());
+                } else {
+                    userList.add(email);
                 }
             }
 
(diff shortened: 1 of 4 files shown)


mcjaeger:
I understood this right that instead of "Freddie Mercury, ..." there will be "freddie.mercury@queen.co.uk"? in the fields in edit and detail view of some document?

mcjaeger:
current screenshot of status in question

screen shot 2017-01-09 at 14 20 37

mcjaeger:
OK, works for non-existing user, sorry, my confusion.

frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/tags/DisplayUserEmail.java
```diff
@@ -33,12 +37,17 @@ public void doTag() throws JspException, IOException {
...
-            user = UserCacheHolder.getUserFromEmail(email);
-        } else {
-            email = "";
+            UserService.Iface client = new ThriftClients().makeUserClient();
+
+            if(client != null) {
+                try {
+                    user = client.getByEmail(email);
+                } catch(TException e) {
+                    log.warn("Exception occurred in User client.");
```
mcjaeger: let's add exception message and parameter email used here ...

frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/tags/DisplayUserEdit.java
```diff
@@ -95,11 +97,13 @@ public int doStartTag() throws JspException {
                     if (!Strings.isNullOrEmpty(email))
                         user = client.getByEmail(email);
                 } catch (TException e) {
-                    user = null;
+                    log.warn("Exception occurred in User client.");
```
mcjaeger: let's add exception message and parameter email used here ...

maximilianhuber:
@mcjaeger the PR is now updated. Could you please update your review?

PR #315, opened 2016-12-21
Show Hierarchy Level at LicenseInfo Generation Page

### LINKED ISSUE #305: Show level of component hierarchy when creating a license information file (ReadmeOSS)
When you have a project which consists of components and also other projects, which themselves contain other projects you get a very big list of components in the license information file generation window and for comfortability and avoiding mistakes it would be good to see the hierarchical level number of each component in the beginning of the selection entry.

alex-evo:
Added column for showing tree level of releases and projects at licenseInfo generation

Closes #305

### DIFF
--- a/backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ProjectDatabaseHandler.java
+++ b/backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ProjectDatabaseHandler.java
@@ -281,7 +281,8 @@ public class ProjectDatabaseHandler {
                         .setRelation(relationship)
                         .setVersion(project.getVersion())
                         .setProjectType(project.getProjectType())
-                        .setClearingState(project.getClearingState());
+                        .setClearingState(project.getClearingState())
+                        .setTreeLevel(visitedIds.size() - 1);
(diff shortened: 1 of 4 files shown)


mcjaeger:
There seems to be an issue with the colspan in the first table header row. (see screenshot) 
Also, my proposal is to change "LVL" to "Level" or at least "Lvl" because "LVL" appears really dominating here. (this is an optical impression issue for me here)

screen shot 2016-12-22 at 14 17 38

mcjaeger:
For the hierarchy, it seems like there is for every node, starting with `0`:

screen shot 2016-12-22 at 14 29 39
(note the indentation)

alex-evo:
That colspan always gets me. Fixed it and changed column title to Lvl.

Are you sure you've deployed cleanly? Because I can't reproduce the problem with level numbering. Looks like this for me:
image

alex-evo:
UPDATE: added showing tree level in attachment rows, where it was missing before

mcjaeger:
works and looks good!

mcjaeger:
@maximilianhuber, @bs-jokri  I think the code changes are very small, would you be OK if take over the review role for this pull request?

PR #320, opened 2017-01-11
Change ECC Fields by Authorized Users Only

alex-evo:
This PR is moves the right and responsibility to change ECC related fields in a release to explicitly authorized users with role "ECC Admin". It also provides ECC Admins with an overview of ECC status of all releases.

- moved ecc fields to a separate tab in display and edit
- added fields assessmentDate and eccStatus
- changed to automatic setting of assessor email, assessor department, and assessment date
- display ecc status in the tab by a colored bubble
- moved ECC fields from clearingInformation to a separate structure
- added handling of eccInformation to release moderation
- added missing ECC fields to Excel export and CSV backup
- added page and portlet for ECC overview of all releases
- added ECC_ADMIN role and corresponding checks for permissions
- only at least ECC_ADMINs may change ECC fields of a release
- only at least ECC_ADMINs can see the ECC overview
- moderation requests for changes in ECC fields are sent to ECC_ADMINs

NOTE: For deployment, there's the new DB migration script 004. Also, a new version of PrivagePages.lar has to be imported.


mcjaeger:

libraries/importers/component-importer/src/main/java/org/eclipse/sw360/importer/ComponentCSVRecord.java
```diff
@@ -288,6 +286,34 @@ private boolean isSetClearingInformation() {
         return false;
     }
 
+    private boolean isSetEccInformation() {
```
maximilianhuber: Why not:
```
return ! (isNullOrEmpty(eccStatus) && 
          isNullOrEmpty(eccAL) &&
          ...)
```
That would be more readable in my opinion

frontend/sw360-portlet/src/main/webapp/html/components/includes/releases/detailOverview.jspf
```diff
@@ -1,18 +1,27 @@
...
                 <li <core_rt:if test="${selectedTab == 'Clearing'}">                     class="active" </core_rt:if> id="Clearing" >   <a href="#tab-ClearingDetails">Clearing Details</a></li>
+                <li <core_rt:if test="${selectedTab == 'ECC Details'}">                  class="active" </core_rt:if> id="ECC-Details"> <a href="#tab-ECCDetails">ECC Details
+                    <div id="eccStatusDiv"
+                        <core_rt:if test="${release.eccInformation.eccStatus.value == 0 || release.eccInformation.eccStatus.value == 3}"> class="notificationBulletSpan backgroundAlert" </core_rt:if> <%--ECCStatus.OPEN || ECCStatus.REJECTED--%>
+                        <core_rt:if test="${release.eccInformation.eccStatus.value == 1}"> class="notificationBulletSpan backgroundWarning" </core_rt:if> <%--ECCStatus.IN_PROGRESS--%>
+                        <core_rt:if test="${release.eccInformation.eccStatus.value == 2}"> class="notificationBulletSpan backgroundOK" </core_rt:if>> <%--ECCStatus.APPROVED--%>
```
maximilianhuber: maybe also add the text `"OK"` to the bullet?

mcjaeger:
will moderator requests be assigned to ECC_ADMIN only from the same group (or department as we call it sometimes)?

alex-evo:
@mcjaeger , moderation requests go to ECC_ADMINS from the same group if there are any, otherwise to all ECC_ADMINS. Ditto for ADMINS.

libraries/importers/component-importer/src/main/java/org/eclipse/sw360/importer/ComponentCSVRecord.java
alex-evo: Yes, but then I'd have to rewrite this style everywhere in this file for consistency. I decided against it and made my changes blend in with the existing code.

libraries/lib-datahandler/src/main/java/org/eclipse/sw360/datahandler/permissions/DocumentPermissions.java
```diff
@@ -78,6 +79,8 @@ protected boolean getStandardPermissions(RequestedAction action) {
             case USERS:
             case CLEARING:
                 return PermissionUtils.isAdmin(user) || isModerator();
+            case WRITE_ECC:
+                return PermissionUtils.isAdmin(user);
```
alex-evo: WRITE_ECC is only applicable to releases. Normally, I'd want to deny everyone WRITE_ECC permission for any object that is not a release, but just to prevent possible future problems, this line ensures that ADMIN may do anything.

libraries/lib-datahandler/src/main/thrift/components.thrift
```diff
@@ -106,15 +113,27 @@ struct COTSDetails{
...
+    3: optional string ECCN, // European control classification number
     4: optional string assessorContactPerson, // email of ECC person
     5: optional string assessorDepartment, // department of ECC person
     6: optional string eccComment, // comments for ecc information
     7: optional string materialIndexNumber, // six digit material index number, string for convenience
+    8: optional string assessmentDate, // Date - YYYY-MM-dd, date of the last editing of ECC information
+}
+struct ClearingInformation {
+    // supplier / ec info
+//    1: optional string AL, // German Ausfuhrliste // moved to EccInformation
```
alex-evo: Not sure what you mean

libraries/lib-datahandler/src/main/thrift/users.thrift
```diff
@@ -17,6 +17,7 @@ enum UserGroup {
     USER = 0,
     ADMIN = 1,
     CLEARING_ADMIN = 2,
+    ECC_ADMIN = 3,
```
alex-evo: Yes, this isn't nice. I think it'll be ok this time if I change to `ECC_ADMIN = 5`. As long as we don't have this code in production, we break any important database by changing the value. What do you think?

maximilianhuber: No, I will do the change, keep your numbering. Your PR is first

libraries/lib-datahandler/src/main/thrift/components.thrift
maximilianhuber: I would change
```
//    1: optional string AL, // German Ausfuhrliste // moved to EccInformation
```
to
```
//    1: optional string AL, // moved to EccInformation
```
for all moved fields.

libraries/lib-datahandler/src/main/java/org/eclipse/sw360/datahandler/permissions/DocumentPermissions.java
maximilianhuber: But aren't the `WRITE_ECC` things exactly the actions which are allowed to ECC_ADMIN and ADMIN? Thus this would be consistend and would work with future ECC information appended to other objects

alex-evo: That's the thing: I didn't want to allow `WRITE_ECC` for other documents until there's a need to. So, I imagined, if ECC fields are added to some other object in the future, whoever adds the fields, will update the permissions. I didn't want to make them standard just now. But then, I can totally see your point. I think we're splitting hairs here. There isn't a measurable difference right now, just in the potential future. Don't you think?

libraries/lib-datahandler/src/main/thrift/components.thrift
alex-evo: yes, good idea

frontend/sw360-portlet/src/main/webapp/html/components/includes/releases/detailOverview.jspf
alex-evo: I tried it. It looks ok, but I'm not sure I want to change it. I think green color by itself is a sufficiently universal indication that it's "OK". If we add text for this status, then we have to think about "In Progress" and "Open", too. And that may just make matters more complicated without bringing clarity for end-user. So, I'd say let's leave it as is.
@mcjaeger, @bs-jokri, what's your take on this? Do you have any preferences?

mcjaeger: my vote: "OK" is OK and In Progress could be "?". But I like to have the bubbles / cartouche without text more.

alex-evo:
rebased and updated. @maximilianhuber, your move

libraries/lib-datahandler/src/main/java/org/eclipse/sw360/datahandler/permissions/DocumentPermissions.java
mcjaeger: @maximilianhuber How about changing this, when we add the functionality to a project / product level?

libraries/lib-datahandler/src/main/thrift/users.thrift
mcjaeger: is Ok then.

libraries/importers/component-importer/src/main/java/org/eclipse/sw360/importer/ComponentCSVRecord.java
mcjaeger: @maximilianhuber see the point but it is also a point oh having it inline with the code before (apart from that I like explicit if statements more for better debuggability when stepping through)

frontend/sw360-portlet/src/main/webapp/html/components/includes/releases/detailOverview.jspf
mcjaeger: @maximilianhuber Ok for having this without text?

libraries/lib-datahandler/src/main/java/org/eclipse/sw360/datahandler/permissions/DocumentPermissions.java
maximilianhuber: **I am OK with the solution given by Alex,** although I think that my change would make it more future proof and more consisted. I prefer to place the logic in the simplest possible location, and move it out later once necessary.

libraries/importers/component-importer/src/main/java/org/eclipse/sw360/importer/ComponentCSVRecord.java
maximilianhuber: **I am OK with that,** but the code is still very ugly and one should refactor it. Nearly every line is duplicated code here, and it is hard to read due to the negations ("if not `eccStatus` is empty return true else if not `eccAl` is empty return true else [....] else return false" is harder to understand than "return ( not all are empty )").

frontend/sw360-portlet/src/main/webapp/html/components/includes/releases/detailOverview.jspf
maximilianhuber: That was just an idea, this can simply be ignored.

mcjaeger:
Sorry there was an UI interface error: first click = no user feedback, second click clicked, but then the second click actually confirmed the merge without comments :-(

So:
review-by:maximilian@huber@tngtech.com

PR #324, opened 2017-01-17
Handling Closed Moderation Requests

### LINKED ISSUE #313: How to handle finished moderation requests
Moderation requests which have been finished (approved or rejected) should not be visible any longer in the list of moderation requests that are still open.
Proposal is to remove them or provide a separate list with completed requests.
This increases the visibility of the open requests and make handling of them easier.

alex-evo:
To improve handling of closed moderation requests and reduce clutter in the Moderation table, the table is split into two tables in separate tabs: one for open moderation requests (NEW and PENDING) and one for closed (APPROVED and REJECTED) ones. Also, Clearing Admins and Admins get the possibility to delete closed moderation requests via a button in the Actions column.

closes #313

### DIFF
--- a/backend/src/src-moderation/src/main/java/org/eclipse/sw360/moderation/db/ModerationDatabaseHandler.java
+++ b/backend/src/src-moderation/src/main/java/org/eclipse/sw360/moderation/db/ModerationDatabaseHandler.java
@@ -1,5 +1,5 @@
 /*
- * Copyright Siemens AG, 2013-2016. Part of the SW360 Portal Project.
+ * Copyright Siemens AG, 2013-2017. Part of the SW360 Portal Project.
  *
  * All rights reserved. This program and the accompanying materials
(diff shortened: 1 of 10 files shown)


mcjaeger:
unfortunately there seems to be a display issue: the menu area does not seem allright with the filtr element above of it. Also the main list are is overlapping with the left elements.

036 oss sw360 20170131 screensshot pull request testing moderatio request

alex-evo:
I'll look into it

alex-evo:
I figured the best way to fix the layout issue was to change some things in the sw360.css and that caused me to realize that the styles for search columns on all pages were not applied (they were lower priority than some aui styles). In the end, I had to change `sw360.css`, `components/view.jsp`, and `projects/view.jsp`.
@mcjaeger, please retest.
@maximilianhuber, please review the new changes.

Issue #326, opened 2017-01-18
Spreadsheet export for Projects on Project Listing


AndreaVogler:
Component type (OSS, COTS, ...) is missing in the Spreadsheet export

The EXCEL export should also contain the included sub projects as separate rows

AndreaVogler:
This is not really related to 'Excel export for Components' !

First topic is clear.

Some more words to the second topic :
Please filter for projects 'Reuse: DNA' and then create an EXCEL report 'Project with linked Releases'.
There are separate rows for each Release, so far so good, this is excatly what the option 'Project with linked Releases' should do.
However there are two contained projects, they are gathered in column T 'linked projects with relation ship'.
Why not spending one line for each of the linked projects as well, then I have a list of all included objects (releases or projects) on this level ?

mcjaeger:
So, the tickets is about two things:
* adding component type to listing (component type is actually at component not at release that is why it is not in the listing)
* adding a third type of selection in the pull down menue: projects with releases including subprojects.

imaykay:
Hi @mcjaeger - if I understood your last comment correctly, this issue is only about adding the component type to each release (=row) in the spreadsheet export "Projects with linked releases" on the "Projects" page, so for the list of all projects and not for the export of a single project (which has been done in #425 only for the clearing status tab).

mcjaeger:
> To solve last case that component_id comes only once would take some extra effort and I thought it might be no problem - what do you think?
yes, I agree

> And if I understood the conversation correctly, the second topic is not part of this issue, right?
yes, this is allright

imaykay:
Alright, then this feature is testable via PR #542

PR #341, opened 2017-02-15
WIP: Added Spreadsheet Export For Specific Products

### LINKED ISSUE #325: Excel export for one specific product
SW360 provides the feature to generate an EXCEL export for the complete or a filtered set of products. This is good but somehow complicated to get the required information for a specific product. There are so many columns and you need to adapt it before you really can work with it, e.g. if you just want to have a short report of the current status which you can send to other people to do some discussions/checks.

I'm missing the feature to create an EXCEL report for one specific product with less information than provided by the general EXCEL report:
- component  name
- version
- type (OSS / COTS or other product)
- clearing state
- security vulnerability
If one of the entries is another product the entry should be expandable to see the next level of components.
See attached example file.

alex-evo:
- added the possibility to download project export for a single project and its subprojects from the "Linked Releases And Projects" tab.
- also renamed Export Excel to Export Spreadsheet

closes #325

### DIFF
--- a/backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ProjectDatabaseHandler.java
+++ b/backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ProjectDatabaseHandler.java
@@ -1,5 +1,5 @@
 /*
- * Copyright Siemens AG, 2013-2016. Part of the SW360 Portal Project.
+ * Copyright Siemens AG, 2013-2017. Part of the SW360 Portal Project.
  *
  * All rights reserved. This program and the accompanying materials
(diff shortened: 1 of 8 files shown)


alex-evo:
I just noticed that I've typed "Stylesheet" instead of "Spreadsheet" 15 times in a row. I'll correct that as soon as I make my other laptop connect to the network.

mcjaeger:
tested, works, but a small idea:

would it possible to also cover the issue #326: adding the component type of the release? The #326 as well as #325 mention the component type (OSS|COTS|INTERNAL|...) which could be added to the spreadsheet output when looking for the component of the release ...

PR #345, opened 2017-02-21
Permission Checks In License Service

### LINKED ISSUE #106: Check create operations in lib-datahandler
Create can be done by anyone, like any user can create components, projects. But a user cannot create licenses. This is done in the portlet level, however, could be also moved to the backend and in the lib-datahandler.

alex-evo:
Added permission checks in license service for creating licenses and related objects where missing.
Deleted executable class `LicenseImporter.java`, which was designed to import license data from command line and has since been replaced by the functionality of `ComponentUploadPortlet`

closes #106 

This PR introduces incompatible changes in Thrift API.
**Increase the minor version after merging**


maximilianhuber:
> **Increase the minor version after merging**

this should be formulated as tag

alex-evo:
1. Agreed
2. I asked about `LicenseImporter.java` in the issue, got no reply, and made a decision on my own. I probably should've mentioned it here, too.

alex-evo:
@maximilianhuber, would you give an "Approved" review, please?

mcjaeger:
ehem, suggestions how to test this are welcome - like I shall write a thrift client? I am testing for "everything behaves normal" then ... / looking at the integration test suite.

alex-evo:
An important thing to test is that importing license data through ComponentUploadPortlet still works. Given that the portlet is only available to admins, it's hard to test the permission checks without writing a thrift client or temporarily changing page permissions to try license import with a normal user.

mcjaeger:
is it that assigned white list selection for a todo is not imported? (which could be fine, just mentioning it if this was intended)

Import from export of another branch (Export Spreadsheet at Clearing Status) seems to work fine.

PR #346, opened 2017-02-21
Added Security Responsibles To Projects

alex-evo:
Added field "Security Responsibles" to projects


libraries/lib-datahandler/src/main/thrift/projects.thrift
```diff
@@ -85,6 +85,7 @@ struct Project {
 //    26: optional set<string> comoderators, //deleted
     27: optional set<string> contributors = [],
     28: optional Visibility visbility = sw360.Visibility.BUISNESSUNIT_AND_MODERATORS,
+    129: optional set<string> securityResponsibles = [],
```
maximilianhuber: I am not sure whether the default value for this optional field is necessary / positive since one can not rely on it.

alex-evo: I'm not sure it's necessary either, but I'm sure it won't hurt.

adrian-evo:
Tested and works fine.

PR #352, opened 2017-03-07
Added Download Spreadsheet button to project's Clearing Status tab

alex-evo:
fix for #325 based on user feedback


mcjaeger:
tested with 2nd level subprojects on testdata
actually subprojects work also on the original functionality

mcjaeger:
suggestions:
* could you add the column origin?
* could you do the rename of the file generated and add project name, version and date of generation, s.th. like "releases-projectname-projectversion-20170309.xlsx"

alex-evo:
Changed the file name as suggested.
Also, while I was at it, refactored the spreadsheet content type into a constant.

PR #358, opened 2017-03-15
chore(mvn): add jgitver support

alexbrdn:


maximilianhuber:
Sadly this does not work. Since the dependencies within SW360 stay at 0.0.0

maximilianhuber:
The problem together with a minimal example is described at https://github.com/jgitver/jgitver-maven-plugin/issues/54

maximilianhuber:
This also breaks the possibility to compile only subprojects

maximilianhuber:
https://github.com/jgitver/jgitver-maven-plugin/issues/40 might be related

maximilianhuber:
A good alternative would be to use a bare `git describe`, which generates an output of the following form:
```
$ git describe
sw360-1.6.0-15-g53211535
```
where
- `sw360-1.6.0` is the most recent tag
- `15` is the number of commits since found tag
- `g53211535` is the commit hash of the current commit.

This might allow us to create versioning for CI.

This does not yet show the branch on which one currently is.

maximilianhuber:
Idea: show the result of `git describe` in the footer but do not set it hard as version. For CI one could use an plugin to generate custom versions

alexbrdn:
@maximilianhuber, this topic appears to be stalled. I suggest that we close the PR for now and reconsider it at a later date. What's your take on it?

maximilianhuber:
Yes you are right. But I think a lightweight release process is already a partial solution.

mcjaeger:
" But I think a lightweight release process is already a partial solution. " -> what would that be (please help)

maximilianhuber:
Our current setup with the semantic versioning already allows for a fast release of a new Minor/Patch version. Thus, if one wants to deploy the latest state and at the same time a tagged state, it is not much overhead to tag the current master.

PR #362, opened 2017-03-30
Lock projects with clearing state CLOSED for editing

### LINKED ISSUE #323: Project with project clearing state 'closed' must not be changed any more
The new property 'project clearing state' has the values : open, in progress, closed.
When a project is in closed state no changes should be allowed anymore for a 'normal' user, e.g. removing or adding releases.

The project then might have already been delivered to customers and also the ReadmeOSS. If a project gets changed after release this causes inconsistencies with already delivered products.

Only experienced users (i.e. clearing admins) might be able to set the state back to 'in progress', do some distinct change and then put it back to 'closed'

This issue has been discussed with Michael Jäger and Marco Wagner on 2017-01-16.

alex-evo:
No user, whether creator, contributor or moderator, is allowed to edit projects in Clearing State CLOSED. Only Clearing Admins and higher can change such projects without triggering a moderation request

closes #323


alex-evo:
oops!

PR #365, opened 2017-04-07
Improved Loading Time of the Project List Page

alex-evo:
- deferred calculation of release clearing state summaries by loading them asynchronously via ajax
- disabled loading of user names by emails for display in the table


mcjaeger:
also tested latest changes according to review.

Issue #236, opened 2016-09-16
License extraction from SPDX XML file not covering depdencies


mcjaeger:
Testing with [SPDXRdfExample-v2.0.rdf.zip](https://github.com/sw360/sw360portal/files/476858/SPDXRdfExample-v2.0.rdf.zip) from the SPDX Tools 2.0.4 revealed that the file dependency with the Apache 2.0 license do not appear in the resulting SPDX output. It should be also there.

mcjaeger:
apparently solved soon by undisclosed source.

mcjaeger:
closed by https://github.com/sw360/sw360portal/pull/370

PR #368, opened 2017-04-13
fix(spdx): `getAllLicenseTexts` was not called due to merge failure

maximilianhuber:
The generated reports does not include all license information due to a merge failure.

### DIFF
--- a/backend/src/src-licenseinfo/src/main/java/org/eclipse/sw360/licenseinfo/parsers/SPDXParser.java
+++ b/backend/src/src-licenseinfo/src/main/java/org/eclipse/sw360/licenseinfo/parsers/SPDXParser.java
@@ -94,7 +94,7 @@ public class SPDXParser extends LicenseInfoParser {
         return new URI("file", filePath, null).toString();
     }
 
-    protected Stream<String> getAllLicenseTextsFromInfo(AnyLicenseInfo spdxLicenseInfo) {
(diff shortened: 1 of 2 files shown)


mcjaeger:
OK, but could you pls. provide some information how to test it? something like take example file YXZ -> now chowing licenses from section ABC ...

maximilianhuber:
The example is `SPDXRdfExample-v2.0.rdf` in which the old version only extracts 5 licenses and the patched code finds 7 (see https://github.com/sw360/sw360portal/pull/368/files#diff-bc64ea2c4db63933f9a7e07f081a9797R68).

I will try to find out which license is affected and why.

maximilianhuber:
@mcjaeger the buggy implementation misses the LGPL in https://raw.githubusercontent.com/spdx/tools/master/TestFiles/SPDXRdfExample-v2.0.rdf.

##

another question: the testfile contains
```
[...]
PackageLicenseInfoFromFiles: GPL-2.0
[...]
LicenseInfoInFile: GPL-2.0
[...]
```
but we currently only look at the concluded licenses. Should we also pay attention to the `licenseInfoInFile` information?

maximilianhuber:
This closes https://github.com/sw360/sw360portal/issues/236

mcjaeger:
@maximilianhuber I think focus concluded licenses and not putting found licenses is needed because licenses found by scanners could be wrong / false positives with significant probability.

mcjaeger:
@maximilianhuber I am sorry I cannot see that Apache is in now in the generated text document (thanks for the link to the issue though)

PR #370, opened 2017-04-13
feat(admin): allow to import all SPDX licenses

maximilianhuber:
This adds a button in the admin db ui which allows an admin to import all SPDX licenses into the DB

### DIFF
--- a/backend/src/src-licenses/src/main/java/org/eclipse/sw360/licenses/tools/SpdxConnector.java
+++ b/backend/src/src-licenses/src/main/java/org/eclipse/sw360/licenses/tools/SpdxConnector.java
@@ -1,0 +1,118 @@
+/*
+ * Copyright (c) Bosch Software Innovations GmbH 2017.
+ * Part of the SW360 Portal Project.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.eclipse.sw360.licenses.tools;
+import org.apache.log4j.Logger;
+import org.eclipse.sw360.datahandler.thrift.SW360Exception;
+import org.eclipse.sw360.datahandler.thrift.licenses.License;
+import org.spdx.compare.LicenseCompareHelper;
+import org.spdx.compare.SpdxCompareException;
+import org.spdx.rdfparser.InvalidSPDXAnalysisException;
+import org.spdx.rdfparser.license.LicenseInfoFactory;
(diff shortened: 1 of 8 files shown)


alex-evo:
@maximilianhuber there's a build error. Probably due to a faulty rebase.

maximilianhuber:
I will have a look.

backend/src/src-licenses/src/main/java/org/eclipse/sw360/licenses/tools/SpdxConnector.java
```diff
@@ -0,0 +1,117 @@
...
+     * @return
+     */
+    public static boolean matchesSpdxLicenseText(License license) {
+        return matchesSpdxLicenseText(license, license.getShortname());
+    }
+
+    public static boolean matchesSpdxLicenseText(License license, String spdxId) {
+        final Optional<SpdxListedLicense> spdxLicense = getSpdxLicense(spdxId);
+        if(!spdxLicense.isPresent()) {
+            return true;
```
alex-evo: Does it mean that any `license` matches its SPDX counterpart when that SPDX counterpart is not found? Looks like a mistake to me.

backend/src/src-licenses/src/main/java/org/eclipse/sw360/licenses/tools/SpdxConnector.java
```diff
@@ -0,0 +1,117 @@
...
+        return  LicenseCompareHelper.isLicenseTextEquivalent(license.getText(), licenseText);
+    }
+
+    /**
+     * returns a list of potentially matching SPDX license IDs
+     * @param license
+     * @return
+     * @throws SW360Exception
+     */
+    public static List<String> findMatchingSpdxLicenseIDs(License license) throws SW360Exception {
```
alex-evo: These last four methods in the class appear to be never used. What's the point?

maximilianhuber: These functions were implemented for a proof-of-concept and might be used in another feature.

backend/src/src-licenses/src/main/java/org/eclipse/sw360/licenses/tools/SpdxConnector.java
maximilianhuber: A license, which has no SPDX counterpart, does **not not** match the SPDX-license.

This function should prevent SPDX-incompatilities, and not only allow SPDX licenses

alex-evo: Okay, I got it. Could you, perhaps, add some code comment to clarify this point inline? Just to accommodate my short memory span.

PR #371, opened 2017-04-13
feat(ui): added custom role map to project, component and release

maximilianhuber:
This PR adds the following features:
* displayed in details-view (as mailto-link) and edit view
* keys can be configured in sw360-portlet: sw360.properties
* edit-view: user can choose key out of drop down list and must enter valid email address per row
* Moderation request generation for role-property
* Diff display of properties of type map of set to set of strings
* merge for properties of type map of set to set of strings

### DIFF
--- a/frontend/sw360-portlet/src/main/java/com/siemens/sw360/portal/tags/DisplayMapOfEmailStringSets.java
+++ b/frontend/sw360-portlet/src/main/java/com/siemens/sw360/portal/tags/DisplayMapOfEmailStringSets.java
@@ -1,0 +1,78 @@
+/*
+ * Copyright (c) Bosch Software Innovations GmbH 2016.
+ * Part of the SW360 Portal Project.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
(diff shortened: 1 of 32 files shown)


frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/common/PortletUtils.java
```diff
@@ -145,8 +146,6 @@ else if (field == Release._Fields.MAINLINE_STATE)
             return getMainlineStatefromString(value);
         else if (field == Project._Fields.STATE)
             return getProjectStateFromString(value);
-        else if (field == Project._Fields.CLEARING_STATE)
```
maximilianhuber: You are right. This looks wrong.

frontend/sw360-portlet/src/main/java/com/siemens/sw360/portal/tags/DisplayMapOfEmailStringSets.java
```diff
@@ -0,0 +1,78 @@
+/*
+ * Copyright (c) Bosch Software Innovations GmbH 2016.
```
maximilianhuber: This code was created 2016.

maximilianhuber:
@alex-evo @mcjaeger: the roles are now updated and the branch is rebased.

mcjaeger:
not so important: actually sw360.properties with old values overwrites default values!

PR #373, opened 2017-04-19
feat(remoteAttachments): minimal implementation

maxhbr:
This PR adds the UI for **adding a remote attachment** to projects, releases, ...:
screenshot

(more information will be added soon)


mcjaeger:
@bs-jokri in case this is active topic, how about discussing it in one of the next telcos

bs-jokri:
@mcjaeger this is still an active topic. However, we currently have some more pressing issues and hence are not working actively on that.

bs-jokri:
As discussed I will close and reopen if we start working on that again.

Issue #287, opened 2016-11-23
Improved Linked Projects and Releases in Project Portlet


mcjaeger:
Edit: please see conversation on what is open.

There are a number of issues now with the display of the table view. Maybe this could be improved for displaying larger projects.

1. add column for type: component (OSS, COTS, Freeware, ...) or proejct type (Internal, Product, Project, Service)
2. Add column for clearing state
3. Presort listing on every level by name.
4. Do not overrun component or project name in the line (so line breaks spoil the display), but shortcut name with '...' after some number of characters, so the line break will not occur. Assuming display width of 1280.
5. Add project attribute project clearing state with values 'open|progress|closed'

mcjaeger:
point 5 is open whether set automatically or manually.

alex-evo:
question: given that generation of license info will get an own page when #280 is merged, should the new columns (Type, Clearing State) be added to that page, too?

AndreaVogler:
add 1: Type (OSS oder COTS) missing in **project page : Clearing Status**

add 3: Sorting in **project page : Linked Releases And Projects** according to release only not by a combination of vendor/release

AndreaVogler:
add 1: Type (OSS oder COTS) missing in project page : Clearing Status

add 3: Sorting in project page : Linked Releases And Projects according to release only not by a combination of vendor/release

Discussed with Michael Jäger / Marco Wagner on 2017-01-16

mcjaeger:
hi, please use @mcjaeger for referring to me so I am getting a notification.

For adding the Type (OSS|COTS|...) we should use short names and not "Open Source Software" and "Commercial". At the sametime, this should be changed also on the main listing of the component listing of the component portlet.

mcjaeger:
still open: 

> For adding the Type (OSS|COTS|...) we should use short names and not "Open Source Software" and "Commercial". At the sametime, this should be changed also on the main listing of the component listing of the component portlet.

mcjaeger:
retested:
* on component entry the short names are used
* in the main listing also shorter naming is used.

PR #376, opened 2017-04-24
Add Component Types INNER_SOURCE and SERVICE and rename existing types

### LINKED ISSUE #350: Component type `service` also for BOM needed.
Modern Apps tend to use services which are provided like openstreetmap, openweather.... the serivces, which are used by an App (i.e. project) shall be listed in sw360, because many of them do have terms of use and OSS involved. Perhaps it is ok to simply add a new `component type  to component which is `Service`.

alex-evo:
- add component types INNER_SOURCE and SERVICE 
- shorten display names of the existing component types

closes #350

### DIFF
--- a/libraries/lib-datahandler/src/main/java/org/eclipse/sw360/datahandler/common/ThriftEnumUtils.java
+++ b/libraries/lib-datahandler/src/main/java/org/eclipse/sw360/datahandler/common/ThriftEnumUtils.java
@@ -1,5 +1,5 @@
 /*
- * Copyright Siemens AG, 2014-2016. Part of the SW360 Portal Project.
+ * Copyright Siemens AG, 2014-2017. Part of the SW360 Portal Project.
  * With modifications by Bosch Software Innovations GmbH, 2016.
  *
  * All rights reserved. This program and the accompanying materials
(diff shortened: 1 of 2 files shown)


maximilianhuber:
The shorter display names might be harder to understand and use. Maybe add hovertexts or descriptions (later)?

libraries/lib-datahandler/src/main/java/org/eclipse/sw360/datahandler/common/ThriftEnumUtils.java
```diff
@@ -48,10 +48,11 @@ private ThriftEnumUtils() {
     );
     //! Enumeration "fancy name" mappings
     private static final ImmutableMap<ComponentType, String> MAP_COMPONENT_TYPE_STRING = ImmutableMap.of(
-            ComponentType.OSS, "Open Source Software" ,
-            ComponentType.COTS, "Commercial Off The Shelf" ,
-            ComponentType.INTERNAL, "Internal" ,
-            ComponentType.FREESOFTWARE, "Free Software");
```
mcjaeger: maybe still 'inner source'

alex-evo: I don't have a preference. I can see how Internal and Inner Source could be more familiar to Siemens employees.
@maximilianhuber , does Bosch have a name for this kind of components (Inner Source)?

mcjaeger:
how about these types?

| before | after |
|---|---|
| OSS  | OSS  |
| COTS  | COTS  |
| Free Software  | Freeware  |
| Internal  | Internal |
| - | Inner Source |
| - | Service |

libraries/lib-datahandler/src/main/java/org/eclipse/sw360/datahandler/common/ThriftEnumUtils.java
maximilianhuber: No, I do not think so

alex-evo:
Updated the component types and texts

mcjaeger:
actually, the tyoe does not seem to be displayed properly in the UI:
1. creating a servcie entry
screen shot 2017-04-28 at 16 18 38
2. displays empty in the listing:
screen shot 2017-04-28 at 16 18 51

alex-evo:
That's a different can of worms. You assigned a type to a component, but you're looking at a release. I know, we talked about this already, but this feature is not yet implemented. That is, the component type is not displayed in the linked releases view. I can't seem find the issue, where this is requested.

mcjaeger:
OK, then let's have this done when solving the issue #287

PR #378, opened 2017-04-24
Restrict project editing to clearing admins of the project group

alex-evo:
dev(projects): restrict project editing to clearing admins of the project group

### DIFF
--- a/backend/src/src-moderation/src/main/java/org/eclipse/sw360/moderation/db/ModerationDatabaseHandler.java
+++ b/backend/src/src-moderation/src/main/java/org/eclipse/sw360/moderation/db/ModerationDatabaseHandler.java
@@ -360,8 +360,8 @@ public class ModerationDatabaseHandler {
                     .collect(Collectors.toList());
         }
 
-        List<User> resultingUsers = relevantUsersOfDepartment.isEmpty() ?
-                (defaultToAllUsersInGroup ? allRelevantUsers : Collections.emptyList()) : relevantUsersOfDepartment;
+        List<User> defaultUsersList = defaultToAllUsersInGroup ? allRelevantUsers : Collections.emptyList();
+        List<User> resultingUsers = relevantUsersOfDepartment.isEmpty() ? defaultUsersList : relevantUsersOfDepartment;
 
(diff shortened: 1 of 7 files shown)


mcjaeger:
we may need a tool tip in the UI, explaining the values

alex-evo:
What values? Are you sure you commented on the right PR?

libraries/lib-datahandler/src/main/java/org/eclipse/sw360/datahandler/permissions/ProjectPermissions.java
```diff
@@ -113,7 +113,7 @@ public boolean isActionAllowed(RequestedAction action) {
             switch (action) {
                 case WRITE:
                 case ATTACHMENTS:
-                    return PermissionUtils.isUserAtLeast(UserGroup.CLEARING_ADMIN, user);
+                    return PermissionUtils.isClearingAdmin(user) && isUserInEquivalentToOwnerGroup() || PermissionUtils.isUserAtLeast(ADMIN, user);
```
maximilianhuber: This is the same as https://github.com/sw360/sw360portal/pull/378/files#diff-cb022dd789f9102c57430a785e4f83c8R83. Maybe encapsulate as function with descriptive name

backend/src/src-moderation/src/main/java/org/eclipse/sw360/moderation/db/ModerationDatabaseHandler.java
```diff
@@ -350,7 +360,8 @@ private String getDepartmentByUserEmail(String userEmail) throws TException {
                     .collect(Collectors.toList());
         }
 
-        List<User> resultingUsers = relevantUsersOfDepartment.isEmpty() ? allRelevantUsers : relevantUsersOfDepartment;
+        List<User> resultingUsers = relevantUsersOfDepartment.isEmpty() ?
+                (defaultToAllUsersInGroup ? allRelevantUsers : Collections.emptyList()) : relevantUsersOfDepartment;
```
maximilianhuber: I think this would be more readable if it would not be inline

PR #381, opened 2017-04-25
Replace compatibility bool with ternary value

maximilianhuber:
This PR replaces the types of two fileds `GPLv2Compatibility` and `GPLv3Compatibility` with a ternary value. This allows to express `"YES"`, `"NO"` and `"UNKNOWN"`.

The corresponding migration script maps the old values in the following way:
- `True` becomes `YES`
- `No` becomes `UNKNOWN`

### DIFF
--- a/frontend/sw360-portlet/src/main/webapp/html/licenses/includes/editDetailSummary.jspf
+++ b/frontend/sw360-portlet/src/main/webapp/html/licenses/includes/editDetailSummary.jspf
@@ -1,5 +1,6 @@
 <%@ page import="org.eclipse.sw360.datahandler.thrift.licenses.LicenseType" %>
 <%@ page import="org.eclipse.sw360.datahandler.thrift.licenses.License" %>
+<%@ page import="org.eclipse.sw360.datahandler.thrift.Ternary" %>
 <%--
   ~ Copyright Siemens AG, 2016. Part of the SW360 Portal Project.
   ~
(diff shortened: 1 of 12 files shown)


mcjaeger:
maybe we merge this with #373 for the increase in version

scripts/migrations/005_convert_compability_fileds_to_ternary.py
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/python
+# -----------------------------------------------------------------------------
+# Copyright (c) Bosch Software Innovations GmbH 2016.
```
alex-evo: 2017?

frontend/sw360-portlet/src/main/webapp/html/licenses/includes/editDetailSummary.jspf
```diff
@@ -50,21 +51,23 @@
...
     <td width="33%">
       <label class="checkboxlabel" for="gpl2">
-        <input id="gpl2" type="checkbox"
-               name="<portlet:namespace/><%=License._Fields.GPLV2_COMPAT%>"
-               value="<sw360:out value="${licenseDetail.GPLv2Compat}"/>"
-                <core_rt:if test="${licenseDetail.GPLv2Compat == 'true'}"> checked="checked" </core_rt:if>
-        />
-        GPL-2.0 compatibility</label>
+        GPL-2.0 compatibility:
+         <select class="toplabelledInput" id="gpl2"
```
alex-evo: Why isn't the new `EditTernary` tag used here? Looks like it isn't used anywhere.

maximilianhuber: `EditTernary` only replaces the inner part of the select. I have removed it since it was incomplete.

libraries/lib-datahandler/src/main/thrift/licenses.thrift
```diff
@@ -86,9 +87,13 @@ struct License {
      9: optional map<string, string> externalIds,
 
     // Additional informations
-	10: optional bool GPLv2Compat,
-	11: optional bool GPLv3Compat,
+	// 10: optional bool GPLv2Compat,
+	// 11: optional bool GPLv3Compat,
 	12: optional string reviewdate,
+
+	15: optional Ternary GPLv2Compat = sw360.Ternary.UNDEFINED,
```
alex-evo: You believe you can skip the qualifying `sw360.` and just write `Ternary.UNDEFINED`

scripts/migrations/005_convert_compability_fileds_to_ternary.py
```diff
@@ -0,0 +1,56 @@
...
+    return compabilityBool
+
+print 'Converting boolean compability flags to ternary.'
+for license_row in licenses:
+    license = license_row.value
+    for field in ['GPLv2Compat','GPLv3Compat']:
+        if field in license:
+            license[field] = convertToTernary(license[field])
+        else:
+            license[filed] = "UNDEFINED"
```
alex-evo: it should be `license[field]`, not `license[filed]`

scripts/migrations/005_convert_compability_fileds_to_ternary.py
```diff
@@ -0,0 +1,56 @@
...
+    if 'issetBitfield' in license:
```
alex-evo: what's this `issetBitfield`?

scripts/migrations/005_convert_compability_fileds_to_ternary.py
alex-evo: Oh, BTW, the filename also has the same typo

alex-evo: And also compatibility is spelled wrong in the filename

scripts/migrations/005_convert_compability_fileds_to_ternary.py
maximilianhuber: In Thrift, if you have some bool value, it is by default `False` and you do not know whether it was actually set. This solves `issetBitfield`.

If `issetBitfiled==0010100...₂` on a binary level, this means that the third and the fifth basic value is set by intent.

Since we remove the last basic value we also remove `issetBitfield` (otherwise the deserializer has problems).

libraries/lib-datahandler/src/main/thrift/licenses.thrift
maximilianhuber: No, this does not work. I do not know why

maximilianhuber:
This branch is now rebased and the requested changes were made.

scripts/migrations/005_convert_compatibility_fields_to_ternary.py
```diff
@@ -0,0 +1,79 @@
...
+    }
+}'''
+
+moderations_with_license_stuff = db.query(moderations_with_license_stuff_fun)
+
+print 'In moderations: converting boolean compatibility flags to ternary.'
+for moderation_row in moderations_with_license_stuff:
+    moderation = moderation_row.value
+    for field in ['licenseAdditions','licenseDeletions']:
+        moderation[field] = map(convertLicenseToTernary, moderation[field])
```
alex-evo: `if field in moderation:` is missing before this line. Otherwise, a KeyError may be thrown

PR #375, opened 2017-04-24
Add project owner, owner accounting unit, and owner group fields

alex-evo:
- Add project owner, owner accounting unit, and owner group fields
- Reorganize display and edit pages for project
- Fix display of project clearing details in project moderation view

PR #382, opened 2017-04-25
Add component owner, owner accounting unit, and owner group fields

alex-evo:
- Add component owner, owner accounting unit, and owner group fields
- Reorganize display and edit pages for component


alex-evo:
I just changed the labels for Owner Accounting Unit and Owner Billing Group to be consistent with #375

PR #385, opened 2017-04-26
chores(rake): fix minor bug in docker call

maximilianhuber:
the previously set volume for the `.m2`-repository-cache was not valid.

(this fixes a feature which is apparently only used by me)

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -48,10 +48,10 @@ task :maybe_build_docker_image do
 end
 
 def runInDocker(dockercmd, moreVolumes="")
-    volumes="-v #{ROOT}:/sw360portal -v _cache/.m2:/root/.m2 #{moreVolumes}"
+    volumes="-v #{ROOT}:/sw360portal #{moreVolumes}"
     workdir="-w /sw360portal"
-    gosu="gosu $(id -u):$(id -g)"
-    sh "docker run -i #{volumes} #{workdir} --net=host sw360/#{DEV_CONTAINER_NAME} #{gosu} #{dockercmd}"
+    chroot="su-exec $(id -u):$(id -g)"
+    sh "docker run -i #{volumes} #{workdir} --net=host sw360/#{DEV_CONTAINER_NAME} #{chroot} #{dockercmd}"
 end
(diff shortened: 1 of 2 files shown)


alex-evo:
I'm curious: what is this feature?

maximilianhuber:
@alex-evo instead of installing Java and Thrift-Compiler locally on your machine you can call
```
rake compile
```
to run a `mavn install` in a dockerized environment.

This also works for deployment and packaging.

alex-evo:
Thanks! I really ought to embrace The Docker Way

alex-evo:
FWIW, `rake compile` fails on my machine with the following error:
```
+ apt-get install -y --no-install-recommends openjdk-8-jdk maven
Reading package lists...
Building dependency tree...
Reading state information...
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 maven : Depends: libaether-java but it is not going to be installed
         Depends: libsisu-ioc-java but it is not going to be installed
 openjdk-8-jdk : Depends: openjdk-8-jre (= 8u121-b13-1~bpo8+1) but it is not going to be installed
                 Depends: openjdk-8-jdk-headless (= 8u121-b13-1~bpo8+1) but it is not going to be installed
E: Unable to correct problems, you have held broken packages.
Removing intermediate container f39f2f92f9e7
The command '/bin/sh -c set -x  && echo "deb http://ftp.debian.org/debian jessie-backports main" > /etc/apt/sources.list.d/backports.list  && $_update && $_install openjdk-8-jdk maven && $_cleanup  && update-alternatives --set java /usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java  && echo "JAVA_HOME=\"/usr/lib/jvm/java-8-openjdk-amd64/jre\"" >> /etc/environment' returned a non-zero code: 100
rake aborted!
Command failed with status (100): [cat sw360dev.Dockerfile | docker build -t ...]

```

maximilianhuber:
@alex-evo  the problem should now be fixed. The container is now based on an image already contianing maven and jdk-8.

PR #398, opened 2017-05-12
feat(export): make lists and maps in spreadsheet export machine-readable

### LINKED ISSUE #353: Spreadsheet Export for Projects should stringify complex objects
For example for releases, the external Id are separated by comma only, having someting like

> key: value, key: value, ...

could be easier to parse when:

> {"key:" "value", "key": "value", ...}

alex-evo:
make lists and maps in spreadsheet export machine-readable by exporting as JSON

closes #353

### DIFF
--- a/libraries/lib-datahandler/src/main/java/org/eclipse/sw360/datahandler/common/SW360Utils.java
+++ b/libraries/lib-datahandler/src/main/java/org/eclipse/sw360/datahandler/common/SW360Utils.java
@@ -8,6 +8,8 @@
  */
 package org.eclipse.sw360.datahandler.common;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.base.Joiner;
 import com.google.common.base.Predicate;
 import com.google.common.base.Strings;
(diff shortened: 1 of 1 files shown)


libraries/lib-datahandler/src/main/java/org/eclipse/sw360/datahandler/common/SW360Utils.java
```diff
@@ -360,20 +362,25 @@ public static String fieldValueAsString(Object fieldValue) {
...
         if (fieldValue instanceof Map) {
-            List<String> mapEntriesAsStrings = nullToEmptyMap(((Map<String, Object>) fieldValue)).entrySet().stream()
-                    .map(e -> {
-                        String valueString = e.getValue() != null ? e.getValue().toString():"";
-                        return e.getKey() + " : " + valueString;
-                    })
-                    .collect(Collectors.toList());
-            return joinStrings(mapEntriesAsStrings);
+            Map<String, Object> originalMap = nullToEmptyMap(((Map<String, Object>) fieldValue));
+            Map<String, String> map = Maps.transformValues(originalMap, v -> v != null ? v.toString() : "");
```
maximilianhuber: the function you have implemented here as lambda is already in the system as `nullToEmptyString`

libraries/lib-datahandler/src/main/java/org/eclipse/sw360/datahandler/common/SW360Utils.java
```diff
@@ -360,20 +362,25 @@ public static String fieldValueAsString(Object fieldValue) {
...
         }
         return fieldValue.toString();
     }
 
+    private static String serializeToJson(Object value) {
+        ObjectMapper mapper = new ObjectMapper();
+        try {
+            return mapper.writeValueAsString(value);
+        } catch (JsonProcessingException e) {
+            throw new IllegalArgumentException(String.format("Cannot serialize field value %s to JSON", value), e);
```
maximilianhuber: It might be better to use no `RuntimeException` here, maybe use instead `SW360Exception`.

maximilianhuber: Please log the exception with `log.error(msg,e)`

libraries/lib-datahandler/src/main/java/org/eclipse/sw360/datahandler/common/SW360Utils.java
```diff
@@ -360,20 +362,25 @@ public static String fieldValueAsString(Object fieldValue) {
...
         }
         if (fieldValue instanceof Iterable){
-            return joinStrings((Iterable<String>) fieldValue);
+            return serializeToJson(fieldValue);
         }
         return fieldValue.toString();
     }
 
+    private static String serializeToJson(Object value) {
+        ObjectMapper mapper = new ObjectMapper();
```
maximilianhuber: It might improve performance to move the `ObjectMapper` to a static field

libraries/lib-datahandler/src/main/java/org/eclipse/sw360/datahandler/common/SW360Utils.java
alex-evo: oops! That's embarrassing. I'll fix it.

libraries/lib-datahandler/src/main/java/org/eclipse/sw360/datahandler/common/SW360Utils.java
alex-evo: It's a seldom used function. I wouldn't care about this performance gain at this point.

libraries/lib-datahandler/src/main/java/org/eclipse/sw360/datahandler/common/SW360Utils.java
alex-evo: You're right.

PR #407, opened 2017-05-16
chore(properties): use default values from java not the properties file

maximilianhuber:
Currently we have the default/fallback to the properties defined in two places:
- in the properties file itself as actual values
- in the java file `PortalConstants.java` which as responsible for parsing the properties.

Current behavior is, that the fallbacks in java are ignored and the value from the properties file are used. The better behavior is to use the java values.

##

This also fixes the issue from comment https://github.com/sw360/sw360portal/pull/371#issuecomment-300535951.


frontend/sw360-portlet/src/main/resources/sw360.properties
```diff
@@ -7,33 +7,33 @@
 # without any warranty.
 #
 
-# where the frontend searches the backend
-backend.url= http://127.0.0.1:8080
+## where the frontend searches the backend
+# backend.url= http://127.0.0.1:8080
```
alex-evo: `ThriftClients` has localhost as default. Please check whether this difference is important

frontend/sw360-portlet/src/main/resources/sw360.properties
```diff
@@ -7,33 +7,33 @@
...
..."Noweb", "NRL", "NTP", "Nunit", "OCLC-2.0", "ODbL-1.0", "PDDL-1.0", "OCCT-PL", "OGTSL", "OLDAP-2.2.2", "OLDAP-1.1", "OLDAP-1.2", "OLDAP-1.3", "OLDAP-1.4", "OLDAP-2.0", "OLDAP-2.0.1", "OLDAP-2.1", "OLDAP-2.2", "OLDAP-2.2.1", "OLDAP-2.3", "OLDAP-2.4", "OLDAP-2.5", "OLDAP-2.6", "OLDAP-2.7", "OLDAP-2.8", "OML", "OPL-1.0", "OSL-1.0", "OSL-1.1", "OSL-2.0", "OSL-2.1", "OSL-3.0", "OpenSSL", "OSET-PL-2.1", "PHP-3.0", "PHP-3.01", "Plexus", "PostgreSQL", "psfrag", "psutils", "Python-2.0", "QPL-1.0", "Qhull", "Rdisc", "RPSL-1.0", "RPL-1.1", "RPL-1.5", "RHeCos-1.1", "RSCPL", "RSA-MD", "Ruby", "SAX-PD", "Saxpath", "SCEA", "SWL", "SMPPL", "Sendmail", "SGI-B-1.0", "SGI-B-1.1", "SGI-B-2.0", "OFL-1.0", "OFL-1.1", "SimPL-2.0", "Sleepycat", "SNIA", "Spencer-86", "Spencer-94", "Spencer-99", "SMLNJ", "SugarCRM-1.1.3", "SISSL", "SISSL-1.2", "SPL-1.0", "Watcom-1.0", "TCL", "Unlicense", "TMate", "TORQUE-1.1", "TOSL", "Unicode-TOU", "UPL-1.0", "NCSA", "Vim", "VOSTROM", "VSL-1.0", "W3C-19980720", "W3C", "Wsuipa", "Xnet", "X11", "Xerox", "XFree86-1.1", "xinetd", "xpp", "XSkat", "YPL-1.0", "YPL-1.1", "Zed", "Zend-2.0", "Zimbra-1.3", "Zimbra-1.4", "Zlib", "zlib-acknowledgement", "ZPL-1.1", "ZPL-2.0", "ZPL-2.1"]
```
alex-evo: Please add comment to this (and possibly other properties) that the value must be a valid JSON array string.

alex-evo:
From testing: default values for Additional Roles contain unnecessary characters.

maximilianhuber:
@alex-evo the mentioned Issue with additional roles is now fixed.

PR #431, opened 2017-06-06
fix(export): improve performance of spreadsheet export and fix simultaneous exports

### LINKED ISSUE #422: Spreadsheet Export Slow or Fails
The export of projects or projects with releases is slow or breaks. tested with 700 projects and their releases. Waiting time is somewhat in 10,20,30 minutes.

Also use of filtering should be tested:
Steps for reproduction:
* Use the tag value in the filter section
* Hit 'Apply'
* Select 'Projects with Linked releases' and hit 'Export Spreadsheet'

Expected:
Export filtered by projects tagged with tag value

alexbrdn:
- use streaming poi api to improve performance of spreadsheet export with many documents
- preload related data in ExcelExporters to reduce roundtrips to DB
- enable simultaneous parallel exports (they were mangled by usage of static fields)
- delete unused field Project.releaseIds

closes #422

### DIFF
--- a/libraries/exporters/src/main/java/org/eclipse/sw360/exporter/ExporterHelper.java
+++ b/libraries/exporters/src/main/java/org/eclipse/sw360/exporter/ExporterHelper.java
@@ -1,5 +1,5 @@
 /*
- * Copyright Siemens AG, 2014-2015. Part of the SW360 Portal Project.
+ * Copyright Siemens AG, 2014-2017. Part of the SW360 Portal Project.
  *
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
@@ -9,9 +9,11 @@
 
 package org.eclipse.sw360.exporter;
 
+import com.google.common.collect.Lists;
 import org.eclipse.sw360.datahandler.thrift.SW360Exception;
 
 import java.util.List;
+import java.util.stream.Collectors;
 
 /**
  * Created by bodet on 06/02/15.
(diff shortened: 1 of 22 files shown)


maximilianhuber:
@alexbrdn 
There is a test that fails:

alexbrdn:
My bad. So much about relying on Travis to run all the tests :/

alexbrdn:
@maximilianhuber , the test's fixed

libraries/exporters/src/main/java/org/eclipse/sw360/exporter/ExporterHelper.java
```diff
@@ -20,10 +22,20 @@
  */
 public interface ExporterHelper<T> {
 
-    public int getColumns();
+    static List<String> addSubheadersWithPrefixesAsNeeded(List<String> headers, List<String> subheaders, String prefix) {
+        List<String> prefixedSubheaders = subheaders
+                .stream()
+                .map(h -> headers.contains(h) ? prefix + h : h)
```
alexbrdn: Releases have substructures like clearingInformation, for example. If I were to prefix all subheaders, we'd get some really long headers like `release: clearing information: componentLicenseInformation`. I didn't like that prospect, so I decided to only prefix duplicate headers.

mcjaeger:
speedwise it is just another dimension: exporting thousands of components (only) is done in 5.6 secs. with releases  not much more resulting in 3.3MB of spreadsheet data. project export of several hundreds, 2.39 secs. wow.

Issue #445, opened 2017-06-09
Re-Evaluate if Linked Projects and Releases and Linkes Releases could be unified.


mcjaeger:
Currently there is some overlap between the linked releases and the linked projects as they show very similar things.

Discussion is, if the two view could be unified to one. Major issue might be response time.

PR #461, opened 2017-06-13
fix(projects): improve project details and project edit pages' performance

alexbrdn:
fix(projects): improve project details and project edit pages' performance

Load linked projects tree and linked releases tree level by level instead of all at once. This improves the loading time of the release details and edit pages and project and release moderation pages as well.


backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ComponentDatabaseHandler.java
```diff
@@ -694,38 +696,38 @@ private void removeReleaseId(String releaseId, String componentId) throws SW360E
...
+                releaseLink.setParentNodeId(parentNodeId);
                 if (release.isSetMainLicenseIds()) {
                     releaseLink.setLicenseIds(release.getMainLicenseIds());
                 }
 
-                if (release.isSetReleaseIdToRelationship()) {
-                    List<ReleaseLink> subreleaseLinks = iterateReleaseRelationShips(release.getReleaseIdToRelationship(), id, visitedIds, releaseMap);
-                    releaseLink.setSubreleases(subreleaseLinks);
-                }
+//                if (release.isSetReleaseIdToRelationship()) {
```
maximilianhuber: commented code?

frontend/sw360-portlet/src/main/webapp/html/projects/includes/linkedProjects.jspf
```diff
@@ -122,8 +72,29 @@
           <portlet:param name="<%=PortalConstants.PAGENAME%>" value="<%=PortalConstants.PAGENAME_SOURCE_CODE_BUNDLE%>"/> \
         </portlet:renderURL>';
     }
+
+    var linkedProjectsTable;
     $(window).load(function () {
-        $("#LinkedProjectsInfo").treetable({expandable: true});
+        linkedProjectsTable = $("#LinkedProjectsInfo");
+        linkedProjectsTable.treetable({
```
maximilianhuber: same code is in `linkedReleaseDetails.jspf`

alexbrdn: That's right. I'm a little torn on whether it would be better to extract a common JS function from this. Logically, it'd ought to go into a new js file, but we make already too many requests to build one page as it is. Putting it into one of the existing js files would be misplacing, I think. Or do you know of a better place to put this code?
Also, there's #445 and maybe both trees will be unified some day.
Do you think I should extract the code nonetheless?

alexbrdn:
I've incorporated most of the review comments.

alexbrdn:
Travis build fails due to #471

frontend/sw360-portlet/src/main/webapp/html/projects/includes/linkedProjects.jspf
maximilianhuber: maybe move the code to some common `.jspf` file which gets imported in both places?

alexbrdn: done

PR #470, opened 2017-06-23
chore(repositories): refactor and cleanup CouchDB Repositories

maxhbr:
This PR tackles multiple issues:
1. Couchdb recommends that Views should not emit the document, i.e. instead of `emit(null, doc)` one should use `emit(null, doc._id)`. This improves performance and reduces the size of the cached views. The documents are still delivered via the `includeDocs=true` property
2. The `ComponentSummary` was inefficiently implemented. If one wanted to have a summary of Components it collected **all Releases for every component** and filtered them later to fill in the details.
  - this raises the question whether Summaries should only be projections?
3. There was no `VendorDatabaseHandler` and there was much logic in the `VendorRepository`.

### DIFF
--- a/backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ComponentDatabaseHandler.java
+++ b/backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ComponentDatabaseHandler.java
@@ -195,10 +195,9 @@ public class ComponentDatabaseHandler extends AttachmentAwareDatabaseHandler {
     }
 
     public List<Component> getMyComponents(String user) {
-        //This call could be sped up, because we want the full documents
-        Set<String> myComponentIds = componentRepository.getMyComponentIds(user);
+        Collection<Component> myComponents = componentRepository.getMyComponents(user);
 
-        return componentRepository.makeSummary(SummaryType.HOME, myComponentIds);
+        return componentRepository.makeSummaryFromFullDocs(SummaryType.HOME, myComponents);
     }
 
     public List<Component> getSummaryForExport() {
@@ -335,7 +334,7 @@ public class ComponentDatabaseHandler extends AttachmentAwareDatabaseHandler {
     }
 
     private boolean isDuplicate(Component component){
-        Set<String> duplicates = componentRepository.getMyComponentIdsByName(component.getName());
+        Set<String> duplicates = componentRepository.getComponentIdsByName(component.getName());
         return duplicates.size()>0;
     }
 
@@ -1118,7 +1117,7 @@ public class ComponentDatabaseHandler extends AttachmentAwareDatabaseHandler {
     public Map<String, List<String>> getDuplicateReleases() {
         ListMultimap<String, String> releaseIdentifierToReleaseId = ArrayListMultimap.create();
 
-        for (Release release : releaseRepository.getAll()) {
+        for (Release release : getAllReleases()) {
             releaseIdentifierToReleaseId.put(SW360Utils.printName(release), release.getId());
         }
 
@@ -1138,7 +1137,7 @@ public class ComponentDatabaseHandler extends AttachmentAwareDatabaseHandler {
     public Map<String,List<String>> getDuplicateReleaseSources() {
         ListMultimap<String, String> releaseIdentifierToReleaseId = ArrayListMultimap.create();
 
-        for (Release release : releaseRepository.getAll()) {
+        for (Release release : getAllReleases()) {
 
             if(release.isSetAttachments()) {
                 for (Attachment attachment : release.getAttachments()) {
--- a/backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ComponentDatabaseHandler.java
+++ b/backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ComponentDatabaseHandler.java
@@ -852,8 +852,12 @@ public class ComponentDatabaseHandler extends AttachmentAwareDatabaseHandler {
         return getLinkedReleases(relations, releaseMap, new ArrayDeque<>());
     }
 
+    public List<Release> getAllReleases() {
+        return releaseRepository.getAll();
+    }
+
     public Map<String, Release> getAllReleasesIdMap() {
-        final List<Release> releases = releaseRepository.getAll();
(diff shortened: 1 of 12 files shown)


maximilianhuber:
@alexbrdn This is the mentioned PR which might improve the performance of Component-related queries.

maximilianhuber:
This travis build fails due to https://github.com/sw360/sw360portal/issues/471

alexbrdn:
On my dev system with 2377 components in the DB, loading time (including rendering) for components page reduced from ~5 s to ~4,2 s. Measurements were in no way precise, but I'm confident in the tendency. Assuming constant time of 2-3 seconds for rendering and executing JS in the browser, that's a good improvement on the backend part.

maximilianhuber:
@mcjaeger: Note, the changes in the CouchDB views make the couchdb harder to debug, since the views do no longer contain the actual documents and instead only consist of a single string which is the ID. This string is not clickable.

backend/src/src-cvesearch/src/main/java/org/eclipse/sw360/cvesearch/datasink/VulnerabilityConnector.java
```diff
@@ -52,14 +53,11 @@ public VulnerabilityConnector() throws IOException{
...
-
-    private Release fillVendor(Release release){
-        if (release != null) {
-            vendorRepository.fillVendor(release);
-        }
-        return release;
+        return Optional.ofNullable(releaseRepository.get(releaseId))
+                .map(r -> {
+                    vendorRepository.fillVendor(r);
+                    return r;
```
alexbrdn: if you make `VendorRepository.fillVendor()` return the filled release, you can make this a one-liner

backend/src/src-vendors/src/main/java/org/eclipse/sw360/vendors/VendorHandler.java
```diff
@@ -34,28 +32,28 @@
  */
 public class VendorHandler implements VendorService.Iface {
 
-    private final VendorRepository vendorRepository;
+    private final VendorDatabaseHandler vendorDatabaseHandler;
```
alexbrdn: I think `handler` would be a good name for this field. Other `Handler`s call their `DatabaseHandler`s like that.

mcjaeger:
hm, at first hand it looks faster, but ... maybe it is not, a few tests:

normal master:
```
Project A 4.62 3.0 2.90
Project B 3.74 3.02 3.05
Project C 12.99
Project D 7.32 7.72
List Components 7.71 7.60 8.82
List Projects 9.13 9.63 9.36
```

With this PR

```
Project A 3.08S
Project B 4.83
Project C 13.97
Project D 7.67
Listing Components 4.43, 4.62, 4.69
Listing Projects  9.83. 10.09 10.64
```

-> not much

mcjaeger:
so, testing again to more confident just doing the components listing with both branches:

test again with master:
651 701 635 665 659
test again with refactor/....
569 553 628 566  594

maxhbr:
related branch: https://github.com/sw360/sw360portal/tree/linked-document-support
related wiki page: https://github.com/sw360/sw360portal/wiki/External-documents-with-CouchDB

maxhbr:
@mcjaeger this PR is now
- rebased and
- corrected (do not use includeDocs for everything)

mcjaeger:
@alexbrdn we would need maybe some review again for the parts changed by Alex

backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ComponentDatabaseHandler.java
```diff
@@ -194,9 +194,9 @@ private void autosetReleaseClearingState(Release releaseAfter, Release releaseBe
 
     public List<Component> getMyComponents(String user) {
         //This call could be sped up, because we want the full documents
```
alexbrdn: Shouldn't this comment be removed, now that you've switched to loading full components in the next line?

backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ComponentDatabaseHandler.java
```diff
@@ -700,8 +700,12 @@ private void removeReleaseId(String releaseId, String componentId) throws SW360E
         return getLinkedReleases(relations, releaseMap, new ArrayDeque<>());
     }
 
+    public List<Release> getAllReleases() {
```
alexbrdn: There are two more calls of `releaseRepository.getAll()` in this class, which should be replaced with calls of this method

backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ComponentRepository.java
```diff
@@ -30,142 +32,122 @@
...
-@View(name = "all", map = "function(doc) { if (doc.type == 'component') emit(null, doc._id) }")
-public class ComponentRepository extends SummaryAwareRepository<Component> {
-
-    private static final String RECENT_VIEW = "function(doc) { if(doc.type == 'component') { emit(doc.createdOn, doc._id) } }";
-
-    private static final String SUBSCRIBERS_VIEW =
-            "function(doc) {" +
+@Views({
+        @View(name = "all",
+                map = "function(doc) { if (doc.type == 'component') emit(null, doc) }"),
```
alexbrdn: Can you please explain the rationale for when you emit `doc._id` and when the complete `doc`?

backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ComponentDatabaseHandler.java
maxhbr: DONE

backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ComponentRepository.java
maxhbr: The goal was to migrate from `doc._id` to `doc` in as many places as possible. That is clearly not possible/good for functions which return a list of IDs. This PR might not be complete, since there are still possibilities for optimizations.

PR #476, opened 2017-06-30
feat(components): restructure components view to show only latest components for performance reasons

alexbrdn:
feat(components): restructure components view to show only latest components for performance reasons
rename Display Filter to Quick Filter and Filters to Advanced Search
use new datatables search api throughout the application
chore(components): reduce view page size by restructuring generated data js code
chore(tables): replace unnecessary hiding of control elements with proper datatables api usage

### DIFF
--- a/frontend/sw360-portlet/src/main/webapp/html/vulnerabilities/view.jsp
+++ b/frontend/sw360-portlet/src/main/webapp/html/vulnerabilities/view.jsp
@@ -35,7 +35,7 @@
         <thead>
         <tr>
             <th class="infoheading">
-                Display Filter
+                Quick Filter
             </th>
         </tr>
         </thead>
@@ -44,9 +44,6 @@
             <td>
                 <input type="text" class="searchbar"
                        id="keywordsearchinput" value="" onkeyup="useSearch('keywordsearchinput')">
-                <br/>
-                <input class="searchbutton" type="button"
-                       name="searchBtn" value="Search" onclick="useSearch('keywordsearchinput')">
             </td>
         </tr>
         </tbody>
@@ -83,7 +80,7 @@
     }
 
     function useSearch(searchFieldId) {
-        vulnerabilityTable.fnFilter( $('#'+searchFieldId).val());
+        vulnerabilityTable.api().search($('#'+searchFieldId).val()).draw();
     }
 
     function createVulnerabilityTable() {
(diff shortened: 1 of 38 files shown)


mcjaeger:
there are display issues under chrome (firefox works)

036 oss sw360 20170630 screenshot feedback on pull request component listing 02

also I think the font is way too large. How about making it balck and 2/3 of the font size of the components

alexbrdn:
Re: display issues: you should clear the browser cache. I'm sure you have the old css in the cache.

mcjaeger:
uhh sorry, that is embarrassing, so many years of web UI tests. works with

Could we have the font a little less eye-catching?

alexbrdn:
yes. will do

maximilianhuber:
Just an idea: Instead of **Components (4083/4083)** I would like to see **Components (showing the last 4083 of 4083)** or maybe **Components (4083 latest of 4083)** without the red color. Or some other hint which helps to find the reason why not all components are listed.

mcjaeger:
so having in the discussion:
* 66% of the size
* black font
* String could be "(4083 newest from 4083)"

alexbrdn:
How about this?
screenshot from 2017-07-05 11 13 34
screenshot from 2017-07-05 11 14 20

maximilianhuber:
that is great!

alexbrdn:
@mcjaeger, @maximilianhuber, @bs-jokri 
I have documented our concerns and decision regarding datatables' server-side processing at https://github.com/sw360/sw360portal/blob/7f57506517a69b8697496e8a1409578953cdca81/docs/architecture/adr-001.md

Any feedback will be appreciated.

mcjaeger:
observed: first log in does not load components - the list is empty. The display shows also 0/something, but it is confusing at first hand.

alexbrdn:
@mcjaeger, the empty list at first login is fixed now

frontend/sw360-portlet/src/main/webapp/html/vulnerabilities/view.jsp
```diff
@@ -35,7 +35,7 @@
         <thead>
         <tr>
             <th class="infoheading">
-                Display Filter
+                Quick Filter
```
bodetc: Should this be done in all tabs?

alexbrdn: Yes, for consistency. Also, #506 will consolidate the snippet to one jspf anyway.

Issue #478, opened 2017-07-05
Licence search is case sensitive


Herschdorfer:
When i search for a license for adding to a component. then this search is case sensitive.

bs-jokri:
Relates to #458

maximilianhuber:
Duplicate of https://github.com/sw360/sw360portal/issues/348

mcjaeger:
But thanks for contributing an issue to us! :)

Herschdorfer:
NP :)

Herschdorfer:
made a fix #479

Issue #458, opened 2017-06-13
License Dialog should search for license short name as well


bs-jokri:
When I try to link a license to a release the license dialog only searches for the long name of licenses. By that it is only possible to search eg for Eclipse ... instead of EPL or GNU instead of GPL
![license dialog](https://user-images.githubusercontent.com/13886571/27073223-65a10d44-5023-11e7-9703-2eb804333be7.png)

Herschdorfer:
we can close this here also

Issue #348, opened 2017-02-23
License search in Component Edit View (Select License) is case sensitive (pictures inside) 


mcjaeger:
Component search is case insensitive and filters the list automatically.

Case-sensitive  search is not very intuitive and inconsistent with other search fields.

![2016-12-07_12_46_47-https___sw360 siemens com_-_components_-_sw360_-_vimperator](https://cloud.githubusercontent.com/assets/3418977/23260277/49fb2442-f9d1-11e6-8b4a-a949e33427b1.png)

![2016-12-07_12_47_21-https___sw360 siemens com_-_components_-_sw360_-_vimperator](https://cloud.githubusercontent.com/assets/3418977/23260263/3fa8e006-f9d1-11e6-88ca-c06048c24e3e.png)

Herschdorfer:
we can close this here as well now.

mcjaeger:
yes, closing

PR #479, opened 2017-07-05
fix(license): make search for licenses case insensitive 

Herschdorfer:
*removed case sensitive license search
*added search also in shortname

related to
#478 #348 #458

### DIFF
--- a/frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/portlets/Sw360Portlet.java
+++ b/frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/portlets/Sw360Portlet.java
@@ -19,6 +19,8 @@ import com.liferay.portal.kernel.servlet.SessionErrors;
 import com.liferay.portal.kernel.servlet.SessionMessages;
 import com.liferay.portal.util.PortalUtil;
 import com.liferay.util.bridges.mvc.MVCPortlet;
+
+import org.apache.commons.lang.StringUtils;
 import org.apache.log4j.Logger;
 import org.apache.thrift.TException;
 import org.eclipse.sw360.datahandler.common.SW360Utils;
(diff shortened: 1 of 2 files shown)


frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/portlets/Sw360Portlet.java
```diff
@@ -198,7 +200,10 @@ public void dealWithLicenseAction(ResourceRequest request, ResourceResponse resp
                         @Override
                         public boolean apply(License input) {
                             String fullname = input.getFullname();
-                            return !isNullOrEmpty(fullname) && fullname.contains(searchText);
+                            String shortname = input.getShortname();
+                            return !isNullOrEmpty(fullname)
```
alexbrdn: `isNullOrEmpty` check is not necessary if you're using `StringUtils.containsIgnoreCase` as it checks for null

frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/portlets/Sw360Portlet.java
```diff
@@ -198,7 +200,10 @@ public void dealWithLicenseAction(ResourceRequest request, ResourceResponse resp
...
+                                   && ((StringUtils.containsIgnoreCase(searchText, fullname)
```
alexbrdn: This searches for `fullname` in `searchText`. It was the other way around before.
I think you should flip the arguments in this and the next line.

Herschdorfer:
your right, pushed the changes

mcjaeger:
yes, squash ... please let us know if help would be helpful. Another thing would be to have a nice commit message ("conventional commits"), like

fix(license): make search for licenses case insensitive

more details here: https://github.com/sw360/sw360portal/wiki/Dev-DoD-and-Style

mcjaeger:
(that all said,  the change to the search is a super useful contribution)

Herschdorfer:
hey,
i combined it and changed the commit message. 
np, will help me also ;)

alexbrdn:
This doesn't work with shortnames :( Bummer, I know.
The reason is that `LicenseSummary` doesn't fill in the field `shortname`. 
This patch solves it:
```diff
index 8a8079d..a8e07ad 100644
--- a/backend/src-common/src/main/java/org/eclipse/sw360/components/summary/LicenseSummary.java
+++ b/backend/src-common/src/main/java/org/eclipse/sw360/components/summary/LicenseSummary.java
@@ -34,6 +34,7 @@ public class LicenseSummary extends DocumentSummary<License> {
                 copyField(document, copy, _Fields.LICENSE_TYPE);
             default:
                 copyField(document, copy, _Fields.ID);
+                copy.setShortname(document.getId());
                 copyField(document, copy, _Fields.FULLNAME);
                 copyField(document, copy, _Fields.LICENSE_TYPE_DATABASE_ID);
         }
```
@Herschdorfer, please update the PR with this one line and we'll be able to merge it.

Herschdorfer:
Would 
`copyField(document, copy, _Fields.SHORTNAME);`

not be the better choice?

alexbrdn:
I don't think so. As it stands currently, a license's shortname _is_ its id (See #135). One cannot guarantee that a license from the database has a shortname at all. Therefore, if we are to use the field shortname in the code, we ought to make sure it is equal to license's id.

That said, there's an idea floating in the room that we should pull shortname and id back apart.

Herschdorfer:
ok pitty...
i changed the commit, we need to have this in mind when changing the id/shortname stuff.

PR #482, opened 2017-07-08
chore(header): add SPDX Identifiers

maximilianhuber:
The Identifiers were added via the following script:
```bash
#!/usr/bin/env bash

set -e

cd $(dirname $0)
cd sw360-sw360portal

firstLineOfEPLHeader="All rights reserved. This program and the accompanying materials"
spdxLicenseIdentifier="SPDX-License-Identifier: EPL-1.0"

SAVEIFS=$IFS
IFS=$(echo -en "\n\b")
for filename in $(find . -type f -not \( -path ./.git -prune \)); do
    if ! grep -q "$spdxLicenseIdentifier" "$filename"; then
        sed -i '/^ \* '"$firstLineOfEPLHeader"'/i \
 * '"$spdxLicenseIdentifier"'\
 *' "$filename"
        sed -i '/^# '"$firstLineOfEPLHeader"'/i \
# '"$spdxLicenseIdentifier"'\
#' "$filename"
        sed -i '/^  ~ '"$firstLineOfEPLHeader"'/i \
  ~ '"$spdxLicenseIdentifier"'\
  ~' "$filename"
        sed -i '/^'"$firstLineOfEPLHeader"'/i \
'"$spdxLicenseIdentifier"'\
' "$filename"
        sed -i '/^  '"$firstLineOfEPLHeader"'/i \
  '"$spdxLicenseIdentifier"'\
' "$filename"
    fi
done

# test:
for filename in $(find . -type f -not \( -path ./.git -prune \)); do
    if grep -q "$firstLineOfEPLHeader" "$filename"; then
        if ! grep -q "$spdxLicenseIdentifier" "$filename"; then
            echo "EPL and no SPDX in: $filename"
        fi
    else
        echo "no EPL in:          $filename"
    fi
done
IFS=$SAVEIFS
```

PR #488, opened 2017-07-10
feat(releases): Just show approved reports for download

lepokle:
The release overview tab inside the components portlets has a column showing all clearing reports as downloads. This commit will hide non-approved reports and enhances the tooltip of approved reports with more information (status, approved by, approved on, approve comment, created by, created on).

Detailed steps:
- modify the download attachment tag to receive the attachment as an object rather than name and id
- add an abstract class for clink generation
- add derived classes for every special case (standard link, bulk link, approved clearing report)
- reworked custom download tags to use new classes and attributes
- fixed unused attribute bare in custom tags for download tags
- add new escaping method in TagUtils since escapeHtml does not escape single quotes (which is necessary because single quotes are used as delimiter for attributes in the jsp templates)
- renamed downloadEnable picture to be consistent with other filenames


frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/tags/DisplayDownloadReport.java
```diff
@@ -1,5 +1,7 @@
 /*
- * Copyright Siemens AG, 2013-2015. Part of the SW360 Portal Project.
+ * Copyright Siemens AG, 2013-2017. Part of the SW360 Portal Project.
+ *
+ * SPDX-License-Identifier: EPL-1.0
```
maximilianhuber: this will cause merge conflicts with https://github.com/sw360/sw360portal/pull/482

frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/tags/DisplayDownloadReport.java
```diff
@@ -8,41 +10,37 @@
...
-    private FluentIterable<Attachment> attachments;
-    private Iterator<Map.Entry<String, String>> attachmentIterator;
+    protected AttachmentType attachmentType;
+    protected Set<Attachment> unfilteredAttachments = Collections.emptySet();
+    protected Iterator<Attachment> attachmentIterator = unfilteredAttachments.iterator();
+
+    @Override
+    protected void prepare() throws JspTagException {
+        this.attachmentIterator = unfilteredAttachments.stream().filter((attachment) -> {
+            return attachmentType == null || attachment.getAttachmentType() == attachmentType;
```
lepokle: From the naming of the class I would say that the class is able to list all attachments regardless of their types. I've seen filter as an additional possibility not as an requirement. That was the reason why I accepted null.
If we always want to filter I would suggest to rename the class to something like 'DisplayAttachmentTypeReport' and mark the attribute as required.

I think both solutions are ok, what do you think?

alexbrdn: Never mind my comment. It's fine as it is. I think I misunderstood the line to mean something else at first. I'll just ascribe it to reading code on Monday morning.

maximilianhuber:
@bodetc please rebase branches to update them:

frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/common/AttachmentPortletUtils.java
```diff
@@ -12,6 +12,8 @@
 import com.liferay.portal.kernel.portlet.PortletResponseUtil;
 import com.liferay.portal.kernel.upload.UploadPortletRequest;
 import com.liferay.portal.util.PortalUtil;
+import org.apache.log4j.Logger;
```
alexbrdn: I take it all back.l I really shouldn't do reviews on Monday mornings.

Issue #467, opened 2017-06-21
Components counter does not update when deleting components


maxhbr:
In component portlet, when deleting components from the list, the small red counter on the upper left in brackets does not update. It only updates after page reload.

PR #490, opened 2017-07-11
fix(ui): Component counter updates automatically after deletion

stevinus1:
The counter number is decremented by 1 and then counter text is set with the new number. This fixes issue #467

### DIFF
--- a/frontend/sw360-portlet/src/main/webapp/html/components/view.jsp
+++ b/frontend/sw360-portlet/src/main/webapp/html/components/view.jsp
@@ -68,7 +68,7 @@
 <div id="header"></div>
 <p class="pageHeader">
     <span class="pageHeaderBigSpan">Components</span>
-    <span class="pageHeaderMediumSpan">(<core_rt:if test="${componentList.size() == totalRows}">${totalRows}</core_rt:if><core_rt:if test="${componentList.size() != totalRows}">${componentList.size()} latest of ${totalRows}</core_rt:if>)</span>
+    <span class="pageHeaderMediumSpan">(<core_rt:if test="${componentList.size() == totalRows}"><span id="componentCounter">${totalRows}</span></core_rt:if><core_rt:if test="${componentList.size() != totalRows}"><span id="componentCounter1">${componentList.size()}</span> latest of <span id="componentCounter2">${totalRows}</span></core_rt:if>)</span>
     <span class="pull-right">
           <input type="button" class="addButton" onclick="window.location.href='<%=addComponentURL%>'"
                 value="Add Component">
@@ -351,6 +351,9 @@
                         success: function (data) {
                             if (data.result == 'SUCCESS') {
                                 componentsTable.row('#' + id).remove().draw(false);
+                                $.each([$('#componentCounter'), $('#componentCounter1'), $('#componentCounter2')], function (i, counter) {
+                                    counter.text(counter.text() - 1);
+                                });
                             }
                             else if (data.result == 'SENT_TO_MODERATOR') {
                                 $.alert("You may not delete the component, but a request was sent to a moderator!");


frontend/sw360-portlet/src/main/webapp/html/components/view.jsp
```diff
@@ -62,7 +62,7 @@
 
 <div id="header"></div>
 <p class="pageHeader">
-    <span class="pageHeaderBigSpan">Components</span> <span class="pageHeaderSmallSpan">(${componentList.size()})</span>
+    <span class="pageHeaderBigSpan">Components</span> <span class="pageHeaderSmallSpan" id="componentCounter">(${componentList.size()})</span>
```
alexbrdn: I believe it would be nicer to have another `<span>` with just the number, without parantheses. Then you wouldn't have to strip the parentheses and add them back. The current way duplicates the rule for displaying the component count. It will create a conflict with #476 which will be hard to reconcile. If you introduce a span for just the number, merging with #476 will be a breeze.

frontend/sw360-portlet/src/main/webapp/html/components/view.jsp
```diff
@@ -291,6 +291,11 @@
                 success: function (data) {
                     if (data.result == 'SUCCESS') {
                         componentsTable.row('#' + id).remove().draw(false);
+                        var oldValue = $('#componentCounter').text();
+                        oldValue = oldValue.substring(1, oldValue.length - 1);
+                        oldValue --;
```
ChristophNiehoff: Maybe at this place it would be nicer to introduce a new variable `newValue` (just for clarity)? Otherwise it is strange to read that the `oldValue` is set...

frontend/sw360-portlet/src/main/webapp/html/components/view.jsp
stevinus1: I have added a new span and taken out the code concerning the parentheses, thanks.

stevinus1:
Waiting for changes on the number display which will eliminate the merge conflict.

mcjaeger:
(from dicussion)
1. let's roolback the X from Y feature when loading the component list
2. let's merge this then after #470

stevinus1:
Edited to decrement both numbers if component counter is shown as a tuple.

frontend/sw360-portlet/src/main/webapp/html/components/view.jsp
```diff
@@ -351,6 +351,15 @@
                         success: function (data) {
                             if (data.result == 'SUCCESS') {
                                 componentsTable.row('#' + id).remove().draw(false);
+                                if (${componentList.size() == totalRows}) {
```
alexbrdn: Does this work? Just by reading code, I have doubts. `totalRows` is not a JS variable, so I wouldn't expect it to be available here. And if it were, this would only work once, because the second time around, the componentList would already be shorter than the initial total number of rows. 
I think what you'd rather want to use is `$('#componentCounter', '#componentCounter1', #'componentCounter2').each(function (counter){ counter.text(counter.text()-1);});`

stevinus1: It works for me in testing (including for doing it more than once), but I can do it the way you suggest?

alexbrdn: OK, after a closer look, I understand why it works. I was a little confused.
Nevertheless, I think the code looks cleaner the way I suggested. Otherwise, you have this weird static `if (true)` or `if (false)` in the generated JS. Also, the actual decrement logig is repeated three times.

stevinus1:
@alexbrdn Have changed the code to iterate over counters with $.each

PR #497, opened 2017-07-13
feat(clearing export): add project origin and component type to exported data

### LINKED ISSUE #425: Spreadsheet export of Clearing Status -> Project Origin Missing
For the spreadsheet export of the clearing status listing of a project, not only the component type (COTS, OSS, Freeware...) is missing, but also the project origin value. Because this listing is flat, this information would be helpful.

related to https://github.com/sw360/sw360portal/issues/326

imaykay:
* added project origin string from gui to export which means that after three project names only "..." will be printed if there are more
* added component type to exported data and therefore added db query for component info for each release (=export row)
* fixed changing releases because of unset attachments - this broke the HashMap keys in releaseStringMap (changing key objects is evil)
* removed unused loggers

fixes #425

### DIFF
--- a/libraries/exporters/src/main/java/org/eclipse/sw360/exporter/ProjectExporter.java
+++ b/libraries/exporters/src/main/java/org/eclipse/sw360/exporter/ProjectExporter.java
@@ -18,17 +18,16 @@ import org.eclipse.sw360.datahandler.thrift.components.Release;
 import org.eclipse.sw360.datahandler.thrift.projects.Project;
 import org.eclipse.sw360.datahandler.thrift.projects.ProjectService;
 import org.eclipse.sw360.datahandler.thrift.users.User;
-import org.apache.log4j.Logger;
 
 import java.util.*;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
-import static org.eclipse.sw360.datahandler.thrift.projects.Project._Fields.*;
+import static org.eclipse.sw360.datahandler.thrift.projects.Project._Fields.DOCUMENT_STATE;
+import static org.eclipse.sw360.datahandler.thrift.projects.Project._Fields.PERMISSIONS;
+import static org.eclipse.sw360.datahandler.thrift.projects.Project._Fields.REVISION;
 
 public class ProjectExporter extends ExcelExporter<Project, ProjectHelper> {
-
-    private static final Logger log = Logger.getLogger(ProjectHelper.class);
 
     private static final Map<String, String> nameToDisplayName;
 
--- a/libraries/exporters/src/main/java/org/eclipse/sw360/exporter/ReleaseHelper.java
+++ b/libraries/exporters/src/main/java/org/eclipse/sw360/exporter/ReleaseHelper.java
@@ -8,11 +8,15 @@
  */
 package org.eclipse.sw360.exporter;
 
+import org.apache.commons.collections4.MapUtils;
 import org.apache.thrift.TException;
+import org.eclipse.sw360.datahandler.common.ThriftEnumUtils;
 import org.eclipse.sw360.datahandler.common.UncheckedSW360Exception;
 import org.eclipse.sw360.datahandler.thrift.ReleaseRelationship;
 import org.eclipse.sw360.datahandler.thrift.SW360Exception;
+import org.eclipse.sw360.datahandler.thrift.ThriftUtils;
 import org.eclipse.sw360.datahandler.thrift.components.*;
+import org.eclipse.sw360.datahandler.thrift.projects.ProjectNamesWithMainlineStatesTuple;
 import org.eclipse.sw360.datahandler.thrift.vendors.Vendor;
 
 import java.util.*;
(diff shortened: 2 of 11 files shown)


libraries/exporters/src/main/java/org/eclipse/sw360/exporter/ReleaseHelper.java
```diff
@@ -74,7 +114,11 @@ private void addFieldValueToRow(List<String> row, Release._Fields field, Release
                 addReleaseIdToRelationShipToRow(release.getReleaseIdToRelationship(), row);
                 break;
             case ATTACHMENTS:
-                row.add(release.attachments.size() + "");
+                String size = "0";
```
alexbrdn: I think `String size = Integer.toString(release.isSetAttachments() ? release.getAttachments().size() : 0);` would be cleaner and would not require implicit casting by `+ "" `

mcjaeger:
I need help with understanding "fixed changing releases because of unset attachments" but apart from that:

works with two projects from productive data

libraries/exporters/src/main/java/org/eclipse/sw360/exporter/ProjectExporter.java
```diff
@@ -80,7 +79,8 @@
     static List<String> HEADERS_EXTENDED_BY_RELEASES = ExporterHelper.addSubheadersWithPrefixesAsNeeded(HEADERS, ReleaseExporter.HEADERS, "release: ");
 
     public ProjectExporter(ComponentService.Iface componentClient, ProjectService.Iface projectClient, User user, List<Project> projects, boolean extendedByReleases) throws SW360Exception {
-        super(new ProjectHelper(projectClient, user, extendedByReleases, new ReleaseHelper(componentClient)));
+        super(new ProjectHelper(projectClient, user, extendedByReleases,
+                new ReleaseHelper(componentClient, false, null, null)));
```
imaykay: Good idea! Updated in new version.

libraries/exporters/src/main/java/org/eclipse/sw360/exporter/ReleaseHelper.java
```diff
@@ -58,6 +77,27 @@ public SubTable makeRows(Release release) throws SW360Exception {
 
     private void addFieldValueToRow(List<String> row, Release._Fields field, Release release) throws SW360Exception {
         switch (field) {
+            case COMPONENT_ID:
+                // first, add data for given field
+                row.add(fieldValueAsString(
+                        release.getFieldValue(org.eclipse.sw360.datahandler.thrift.components.Release._Fields.ID)));
```
imaykay: Glad you saw this one! Thanks!

libraries/exporters/src/main/java/org/eclipse/sw360/exporter/ReleaseHelper.java
```diff
@@ -58,6 +77,27 @@ public SubTable makeRows(Release release) throws SW360Exception {
...
+
+                // second, add joined data - but only if wanted
+                // remark that headers have already been added accordingly
+                if (extendedByAdditionalData) {
+                    try {
+                        Component component = cClient.getComponentById(release.componentId, this.user);
```
imaykay: Okay, I added some batch loading for components of releases and of linked releases. Hope that this does not slow the performance too much now.

libraries/exporters/src/main/java/org/eclipse/sw360/exporter/ReleaseHelper.java
imaykay: You are right - I just did the minimal rework to keep it working and kept the implicit cast... Updated in new version.

libraries/exporters/src/main/java/org/eclipse/sw360/exporter/ReleaseHelper.java
```diff
@@ -41,14 +79,11 @@ public int getColumns() {
...
     @Override
     public List<String> getHeaders() {
-        return HEADERS;
+        return addAdditionalData() ? HEADERS_EXTENDED_BY_ADDITIONAL_DATA : HEADERS;
     }
 
     @Override
     public SubTable makeRows(Release release) throws SW360Exception {
-        if (!release.isSetAttachments()) {
-            release.setAttachments(Collections.emptySet());
```
imaykay: `release` is key of the map `releaseToShortenedStringsMap` which is why it should not be modified here. Otherwise the keys won't work anymore cause the hashCode changes after the object has been changed and the value assigned to the key cannot be identified from now on.

imaykay:
@alexbrdn please re-review the requested changes :-)

@mcjaeger I just added a comment to the according code line in `libraries/exporters/src/main/java/org/eclipse/sw360/exporter/ReleaseHelper.java` in the "files changed" tab here to explain the implementation detail I had to rework and which I shortly described by "fixed changing releases because of unset attachments"

libraries/exporters/src/main/java/org/eclipse/sw360/exporter/ReleaseHelper.java
```diff
@@ -58,6 +93,36 @@ public SubTable makeRows(Release release) throws SW360Exception {
...
+
+                // second, add joined data - but only if wanted
+                // remark that headers have already been added accordingly
+                if (addAdditionalData()) {
+                    // add component type
+                    try {
+                        Component component = this.preloadedComponents.get(release.componentId);
+                        if (component != null) {
+                            row.add(fieldValueAsString(component.getFieldValue(
+                                    org.eclipse.sw360.datahandler.thrift.components.Component._Fields.COMPONENT_TYPE)));
```
alexbrdn: `component.getComponentType()`  is easier to read, I would say. Same goes for line 99

libraries/exporters/src/main/java/org/eclipse/sw360/exporter/ReleaseHelper.java
```diff
@@ -58,6 +93,36 @@ public SubTable makeRows(Release release) throws SW360Exception {
...
+                        } else {
+                            row.add("");
+                        }
+                    } catch (TException e) {
```
alexbrdn: I don't think this `try-catch` block is necessary. None of the method calls inside seem to throw `TException`

libraries/exporters/src/main/java/org/eclipse/sw360/exporter/ReleaseHelper.java
imaykay: yeah, right :-)

libraries/exporters/src/main/java/org/eclipse/sw360/exporter/ReleaseHelper.java
imaykay: fieldValueAsString throws SW360Exception which was why my IDE did not show unnecessary try-catch. But without even that method, it clearly should be removed.

mcjaeger:
unfortunately we have found that the export with releases of the project listing runs now for minutes (again) with 750 projects ... but it was in master already

PR #505, opened 2017-07-17
feat(ui): Added the project clearing state to the listing of projects…

ChristophNiehoff:
…. Display project state and project clearing state by coloured boxes. Added the project state to the listings of linked projects and releases.

### DIFF
--- a/backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ProjectDatabaseHandler.java
+++ b/backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ProjectDatabaseHandler.java
@@ -315,6 +315,7 @@ public class ProjectDatabaseHandler {
                         .setParentNodeId(parentNodeId)
                         .setRelation(relationship)
                         .setVersion(project.getVersion())
+                        .setState(project.getState())
                         .setProjectType(project.getProjectType())
                         .setClearingState(project.getClearingState())
                         .setTreeLevel(visitedIds.size() - 1);
--- a/backend/src/src-components/src/test/java/org/eclipse/sw360/components/db/ProjectDatabaseHandlerTest.java
+++ b/backend/src/src-components/src/test/java/org/eclipse/sw360/components/db/ProjectDatabaseHandlerTest.java
@@ -170,6 +170,7 @@ public class ProjectDatabaseHandlerTest {
                 .setNodeId("P3")
                 .setParentNodeId("P2")
                 .setProjectType(ProjectType.CUSTOMER)
+                .setState(ProjectState.ACTIVE)
                 .setTreeLevel(2)
                 .setLinkedReleases(Arrays.asList(releaseLinkR2A, releaseLinkR2B))
                 .setSubprojects(Collections.emptyList());
@@ -178,6 +179,7 @@ public class ProjectDatabaseHandlerTest {
                 .setNodeId("P4")
                 .setParentNodeId("P2")
                 .setProjectType(ProjectType.CUSTOMER)
+                .setState(ProjectState.ACTIVE)
                 .setTreeLevel(2)
                 .setSubprojects(Collections.emptyList());
         ProjectLink link2 = new ProjectLink("P2", "project2")
@@ -185,6 +187,7 @@ public class ProjectDatabaseHandlerTest {
                 .setNodeId("P2")
                 .setParentNodeId("P1")
                 .setProjectType(ProjectType.CUSTOMER)
+                .setState(ProjectState.ACTIVE)
                 .setTreeLevel(1)
                 .setLinkedReleases(Arrays.asList(releaseLinkR1A, releaseLinkR1B))
                 .setSubprojects(Arrays.asList(link3, link4));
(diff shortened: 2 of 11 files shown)


ChristophNiehoff:
The listing of projects looks like this:

project001

In the listings of linked projects and releases the project state is now also shown:

project002

Design ok?

frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/tags/DisplayEnumBox.java
```diff
@@ -0,0 +1,39 @@
...
+    public void setValue(TEnum value) {
+        this.value = value;
+    }
+
+    public String makeBox(String state) {
+        String projectStateBackgroundColour = "backgroundGrey";
+        if( state == "Active" ) {
+            projectStateBackgroundColour = "backgroundOK";
+        }
+        return "<div class=\"stateBox roundEdgesLeft roundEdgesRight " + projectStateBackgroundColour + "\" title=\"Project state: " + state + "\"> PS </div>";
```
alexbrdn: I'm confused why the class is named `DisplayEnumBox` and it takes any `TEnum` value to display, but actually produces only project clearing state box.

ChristophNiehoff: The reason is that I just made a  copy of the class `DisplayEnum` .  But you are right, the name and the type should be changed.

frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/tags/DisplayEnumBox.java
```diff
@@ -0,0 +1,39 @@
...
+
+    private TEnum value;
+
+    public void setValue(TEnum value) {
+        this.value = value;
+    }
+
+    public String makeBox(String state) {
+        String projectStateBackgroundColour = "backgroundGrey";
+        if( state == "Active" ) {
```
alexbrdn: If you are going to change the type, I think a direct comparison of `value == ProjectState.ACTIVE` would be more reliable than string comparison here. Otherwise, this code will break if someone ever changes the string representation of `ProjectState.ACTIVE` in `ThriftEnumUtils`.

mcjaeger:
Two things:
* the CS box is always grey, regardless which project clearing state value is set (open, in progress, closed)
* loading the project list number items per entry results in a "failed to load" for the numbers (i see them in the screenshots herein this conversation, acknowledged) with following exception:
```
Jul 18, 2017 10:00:42 PM org.apache.catalina.core.ApplicationDispatcher invoke
SEVERE: Servlet.service() for servlet projects Servlet threw exception
java.lang.NullPointerException
	at org.eclipse.sw360.portal.portlets.projects.ProjectPortlet.serveGetClearingStateSummaries(ProjectPortlet.java:238)
	at org.eclipse.sw360.portal.portlets.projects.ProjectPortlet.serveResource(ProjectPortlet.java:145)
	at com.liferay.portlet.FilterChainImpl.doFilter(FilterChainImpl.java:122)
	at com.liferay.portlet.ScriptDataPortletFilter.doFilter(ScriptDataPortletFilter.java:82)
	at com.liferay.portlet.FilterChainImpl.doFilter(FilterChainImpl.java:116)
	at com.liferay.portal.kernel.portlet.PortletFilterUtil.doFilter(PortletFilterUtil.java:71)
	at com.liferay.portal.kernel.servlet.PortletServlet.service(PortletServlet.java:112)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:731)
```

ChristophNiehoff:
Implemented the change requests by @alexbrdn and cleaned up the CSS (as suggested by @maximilianhuber )

@mcjaeger : As already said in the discussion, both of the problems you found are connected. Project clearing state and release clearing status are loaded at the same time. I think the problem stems from loading the project clearing state. If that fails then also the loading of the other numbers fails.
I now changed the code such that if the project clearing state is returned as `null` then the other thinks run through. Could you check if after a clean rebuild the incorrect behaviour still shows up?  Does it also appear if there are several projects in the list?  Does it change if you manually edit the `clearingState` field of the project in the couchDB?

mcjaeger:
I works on a clean instance. Will try with staging now.

mcjaeger:
I think we need to debug together. all I can observe is that is works for a clean instance with small example data created. But on our stage server it is not working. Now I observe endless (> 4mins) "Loading ..." and CS box stays grey.

ChristophNiehoff:
After the debug session the code should work now on a larger data sample. @mcjaeger : Could you retest it?

mcjaeger:
works.

alexbrdn:
@ChristophNiehoff, it didn't occur to me at the first reading, but now I see a little problem with the code and can't figure out if there is a legitimate reason for it. Why do you load these two fields with the ajax at all? What's preventing you from generating the table with these fields right from the start? The Clearing State column is loaded asyncronously, because calculating the values for it takes a long time, but I see no such problem with the two state fields. They are just properties of a project after all

mcjaeger:
I fully think it is a valid point of @alexbrdn my idea is just to merge this now and apply a correction. Because we reached a state where it is working.

PR #508, opened 2017-07-19
refactor(attachments list): new look of list view for attachments

### LINKED ISSUE #454: List view for attachments needs rework
The list view needs improvements:

* it has odd use of space,
* multi-row, multi column view looks difficult to catch important information,
* no sorting possible
* some information (approver comment) could be displayed at mouse over only?
* File size is missing.
* approved attachments shall have some indications (use of colour?) to let the user quickly see approved attachments

imaykay:
* removed old attachmentsDetail.jsp code including its css in sw360.css
* added new attachment list view based on jquery datatable plugin with child rows
** added data generation as writing javascript array into the html via jsp
** added js code to generate table from data and dynamically add child rows for details
** added css partly to jsp (different size adjustments) and mainly to sw360.css
** added two new tags for type shortnames and too long text with ellipses
** added idea of shortnames for enum types to ThriftEnumUtils
* not yet done - needs further ticket:
** size and usage data, just n/a is printed in table cell
** edit and delete actions

closes #454

### DIFF
--- a/frontend/sw360-portlet/src/main/webapp/html/utils/includes/attachmentsDetail.jsp
+++ b/frontend/sw360-portlet/src/main/webapp/html/utils/includes/attachmentsDetail.jsp
@@ -8,95 +8,169 @@
   ~ which accompanies this distribution, and is available at
   ~ http://www.eclipse.org/legal/epl-v10.html
   --%>
-<%@ taglib prefix="c" uri="http://java.sun.com/jsp/jstl/core" %>
-<%@include file="/html/init.jsp" %>
 
+<%@ page import="org.eclipse.sw360.portal.common.PortalConstants"%>
 
+<%@ taglib prefix="c" uri="http://java.sun.com/jsp/jstl/core"%>
 
-<portlet:defineObjects/>
-<liferay-theme:defineObjects/>
+<%@include file="/html/init.jsp"%>
+
+<portlet:defineObjects />
+<liferay-theme:defineObjects />
 
 <c:catch var="attributeNotFoundException">
-    <jsp:useBean id="attachments" type="java.util.Set<org.eclipse.sw360.datahandler.thrift.attachments.Attachment>" scope="request"/>
-    <jsp:useBean id="documentType" type="java.lang.String" scope="request"/>
-    <jsp:useBean id="documentID" class="java.lang.String" scope="request"/>
+    <jsp:useBean id="attachments" type="java.util.Set<org.eclipse.sw360.datahandler.thrift.attachments.Attachment>" scope="request" />
+    <jsp:useBean id="documentType" type="java.lang.String" scope="request" />
+    <jsp:useBean id="documentID" class="java.lang.String" scope="request" />
 </c:catch>
+
 <core_rt:if test="${empty attributeNotFoundException}">
-<table class="table info_table " id="attachmentDetail" title="Attachment Information">
-    <thead>
-    <tr>
-        <th colspan="8" class="headlabel">
-            Attachments
-            <core_rt:if test="${not empty attachments}">
-                <sw360:DisplayDownloadAttachmentBundle name="AttachmentBundle.zip"
-													   attachments="${attachments}"
-                                                       contextType="${documentType}"
-                                                       contextId="${documentID}"/>
-            </core_rt:if>
-        </th>
-    </tr>
-    </thead>
-    <tbody>
-    <core_rt:if test="${not empty attachments}">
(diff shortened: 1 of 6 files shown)


libraries/lib-datahandler/src/main/java/org/eclipse/sw360/datahandler/common/ThriftEnumUtils.java
```diff
@@ -83,6 +83,29 @@ private ThriftEnumUtils() {
             .put(AttachmentType.OTHER, "Other")
             .build();
 
+    // @formatter:off
+    public static final ImmutableMap<AttachmentType, String>
+            MAP_ATTACHMENT_TYPE_SHORT_STRING = ImmutableMap.<AttachmentType, String>builder()
+            .put(AttachmentType.DOCUMENT, "DOC")
+            .put(AttachmentType.SOURCE, "SRC")
+            .put(AttachmentType.DESIGN, "DDT")
```
mcjaeger: just a very picky remark: how about changing DDT to DSN

imaykay: Sure, no problem =)

imaykay:
@mcjaeger regarding your requested changes:

1. Table header "checked by" -> "approved by": the column contains also "rejected by" users which only differs in font color. That was the reason for me to rename it to "checked by" instead of "approved by". If you still favor "approved by", it is no problem at all to rename it.

2. "Team" -> "Group": Almost done ;-)

3. Place group column after "... by" column: as said in the feature ticket it then might cause problems because the group of the uploaded-by e-mail address is closer to the approved-by e-mail address when the first one is rather short. If the columns are switched like they are now the group is in every case close to the e-mail address it belongs to. But like in 1, I can change it anyway if you would still like it to be changed.

frontend/sw360-portlet/src/main/webapp/html/utils/includes/attachmentsDetail.jsp
```diff
@@ -21,82 +23,155 @@
...
-
-    <core_rt:if test="${empty attachments}">
-        <tr id="noAttachmentsRow">
-            <td colspan="6">No attachments yet.</td>
-        </tr>
-    </core_rt:if>
-    </tbody>
-</table>
+
+	<core_rt:if test="${empty attachments}">
```
alexbrdn: There's a lot of mixed indenting by tabs and spaces in the file. Please use 4 spaces for indenting.

frontend/sw360-portlet/src/main/webapp/html/utils/includes/attachmentsDetail.jsp
```diff
@@ -21,82 +23,155 @@
...
+
+			/* Define function for child row creation, which will contain additional data for a clicked table row */
+			function createChildRow(rowData) {
+				var childHtmlString = '' +
+						'<div>' +
+							'<span class="dataTableChildRowCell" style="padding-right: 10px; width:  4%;"/>' +
+							'<span class="dataTableChildRowCell" style="padding-right: 50px; width: 36%;">' + rowData.sha1 + '</span>' +
+							'<span class="dataTableChildRowCell" style="padding-right: 30px; width: 22%;">' + rowData.uploadedOn + ' ' + rowData.uploadedComment + '</span>';
+				if (rowData.checkStatus === 'ACCEPTED') {
+					childHtmlString += '' +
```
alexbrdn: can you explain to me what's the purpose of concatenating with this empty string `'' +` all the time, please?

imaykay: This is just to get a clear view on the generated html with its correct indentations for nesting, starting at the normal JS indent for a new line at the opening of the html structure to create.
I could also delete the empty strings, pull the first html line up and indent every following line much more. My main propose is to see the nesting of the generated html code - no matter at what "base indentation" level this happens :-)

frontend/sw360-portlet/src/main/webapp/html/utils/includes/attachmentsDetail.jsp
imaykay: fixed - my editor settings as well!

frontend/sw360-portlet/src/main/webapp/html/utils/includes/attachmentsDetail.jsp
alexbrdn: Got it. Neat.

PR #510, opened 2017-07-20
chore(header): replace minimal headers with the epl related option

maximilianhuber:
Made mainly with the following `sed` script:
```
/\# Copying and distribution of this file, with or without modification/,/\# without any warranty./c\
# All rights reserved. This program and the accompanying materials are made\
# available under the terms of the Eclipse Distribution License v1.0 which\
# accompanies this distribution, and is available at\
# http://www.eclipse.org/org/documents/edl-v10.php
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,10 +3,10 @@
 # Copyright Bosch Software Innovations GmbH, 2017.
 # Part of the SW360 Portal Project.
 #
-# Copying and distribution of this file, with or without modification,
-# are permitted in any medium without royalty provided the copyright
-# notice and this notice are preserved. This file is offered as-is,
-# without any warranty.
+# All rights reserved. This configuration file is provided to you under the
+# terms and conditions of the Eclipse Distribution License v1.0 which
+# accompanies this distribution, and is available at
+# http://www.eclipse.org/org/documents/edl-v10.php
 #
 # initial author: maximilian.huber@tngtech.com
 # -----------------------------------------------------------------------------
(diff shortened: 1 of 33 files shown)


mcjaeger:
I think it does not make sense for some of the file to be put under EPL-1.0, for example the notice file: why do we put the notice file under GPL? I am not sure if this is correct.

bs-jokri:
Actually it is not the EPL but the EDL (Eclipse Documentation License, http://www.eclipse.org/org/documents/edl-v10.php), which is a BSD Style license and the most permissive Eclipse foundation accepts.

NOTICE-distribution
```diff
@@ -331,10 +331,10 @@ THE SOFTWARE.
...
 --------------------------------------------------------------------------------
 
-Copying and distribution of this file, with or without modification,
-are permitted in any medium without royalty provided the copyright
-notice and this notice are preserved.  This file is offered as-is,
-without any warranty.
+All rights reserved. This program and the accompanying materials are made
+available under the terms of the Eclipse Distribution License v1.0 which
+accompanies this distribution, and is available at
+http://www.eclipse.org/org/documents/edl-v10.php
```
mcjaeger: maybe false hit of the script, pls consider removing.

maximilianhuber:
The used sed-script was
```
/\# Copying and distribution of this file, with or without modification/,/\# without any warranty./c\
# All rights reserved. This configuration file is provided to you under the\
# terms and conditions of the Eclipse Distribution License v1.0 which\
# accompanies this distribution, and is available at\
# http://www.eclipse.org/org/documents/edl-v10.php
```
and applied via
```
$ find * -type f -exec sed -i -f ~/TNG/SCT/SW360/refactorHeader1.sed {} \;
```

Issue #335, opened 2017-02-14
Spreadsheet export : one row for each contained / linked project


AndreaVogler:
An EXCEL report with the option 'Project with linked releases' for each release creates one row for each release in the EXCEL sheet.
All linked project are listed in **one** column separated by comma.

Please handle linked projects in the same way like linked releases and create in the  EXCEL report one row for each linked project as well.

mcjaeger:
yes, whould be two stage denormalized.

mcjaeger:
problem is that sub projects and releases are orthogonal

mcjaeger:
continued therehttps://github.com/sw360/sw360portal/issues/435
closing here

PR #512, opened 2017-07-21
fix(ui): Changing content of vendorNames to be vendor short names

### LINKED ISSUE #355: Use vendor short name for listings
Vendors have vendor short name and vendor long name. For the listing there is the long name confusing: for example "Apache Software Foundation", where the "Apache" would be enough.

stevinus1:
Showing vendor short name rather than vendor long name in component listings by changing the content of the vendorNames field,

closes #355

### DIFF
--- a/libraries/importers/component-importer/src/main/java/org/eclipse/sw360/importer/ComponentImportUtils.java
+++ b/libraries/importers/component-importer/src/main/java/org/eclipse/sw360/importer/ComponentImportUtils.java
@@ -469,9 +469,21 @@ public class ComponentImportUtils {
     private static Map<String, String> getVendorNameToVendorId(VendorService.Iface vendorClient) throws TException {
         Map<String, String> vendorNameToVendorId = new HashMap<>();
 
-        for (Vendor vendor : vendorClient.getAllVendors()) {
-            vendorNameToVendorId.put(vendor.getFullname(), vendor.getId());
-        }
+            for (Vendor vendor : vendorClient.getAllVendors()) {
+                if (!vendorNameToVendorId.containsKey(vendor.getShortname())) {
+                    vendorNameToVendorId.put(vendor.getShortname(), vendor.getId());
+                }
+                else {
+                    log.error("There is a clash between the shortname " + vendor.getShortname() + " and a name already present in the mapping.");
+                }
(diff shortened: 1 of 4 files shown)


mcjaeger:
seems to work for the project part, but .. I see that issue #335 was not well described ... could you pls. consider changing that in components as well?

attached an example for the main listing of components where the long name (=apache software foundation) instead of the short name (=apache) is used, but the short name would be a better fit here.

screen shot 2017-07-25 at 17 17 42

libraries/importers/component-importer/src/main/java/org/eclipse/sw360/importer/ComponentImportUtils.java
```diff
@@ -469,9 +469,21 @@ public ReleaseLinkCSVRecord apply(CSVRecord input) {
...
         Map<String, String> vendorNameToVendorId = new HashMap<>();
 
-        for (Vendor vendor : vendorClient.getAllVendors()) {
-            vendorNameToVendorId.put(vendor.getFullname(), vendor.getId());
-        }
+            for (Vendor vendor : vendorClient.getAllVendors()) {
+                if (!vendorNameToVendorId.containsKey(vendor.getShortname())) {
+                    vendorNameToVendorId.put(vendor.getShortname(), vendor.getId());
+                }
+                else {
```
alexbrdn: That's an unusual formatting. The convention is to place `else` on the same line as the previous closing curly brace. Like this: `} else {`.
Same goes for line 482

libraries/importers/component-importer/src/main/java/org/eclipse/sw360/importer/ComponentImportUtils.java
```diff
@@ -469,9 +469,21 @@ public ReleaseLinkCSVRecord apply(CSVRecord input) {
...
+                    log.error("There is a clash between the shortname " + vendor.getShortname() + " and a name already present in the mapping.");
```
alexbrdn: Is it really an error, if you just ignore it and proceed anyway? Maybe you want to log it as a warning?

But that's more of a philosophical question than a call to action, because this module for components csv export should be removed altogether. It's a dead feature. It's slow to the point of being unusable with large amounts of data and no one uses it anyways.

PR #520, opened 2017-07-26
fix(ui): Corrected clearing state display in projects list. The clear…

### LINKED ISSUE #519: Correct Project Clearing State Display in Project List
During the PR https://github.com/sw360/sw360portal/pull/505 we have had the discussion why having the project clearing state actually loaded via the ajax call that calls the statistics. In fact, it could be loaded directly because currently it would be a flat attribute maybe (see discussion in the PR initiated by @alexbrdn), despite it is related with the numbers.

ChristophNiehoff:
…ing state is now loaded directly (not by an ajax call).

closes #519


alexbrdn:
this line can be deleted, too: https://github.com/sw360/sw360portal/pull/520/files#diff-8608ca9de95f951c8008d88501c2b99cR301

PR #521, opened 2017-07-26
feat(ui): More information for releases search

ChristophNiehoff:
- Added clearing state and mainline state to the release search
 - Changed the structure of the search table to `DataTable`, therefore improved the layout and made the columns sortable
 - Merged the first two search buttons (search by name prefix and search by vendor) into a single one, but the last button (Releases of linked projects) is kept.


ChristophNiehoff:
* The search dialog for linking releases to a project now contains the release clearing state and mainline state.

* The search function is modified. Before there were three buttons: search-by-name, search-by-vendor, releases-of-linked-projects. The first two searches are merged into a single button. Therefore, performing a search now results in releases that contain the searched string either in their name or in their vendor's name (short or long name). 
The third button (releases-of-linked-projects) does not refer to the search-string and just returns all releases from linked projects. So, I think it makes more sense to keep it as a seperate button. Agreed?

* To improve the formatting of the table and make it sortable, the table technology is switched to DataTables. Input on my javascript implementation for generating the table is appreciated.  
  The project-edit-page also contains a similar search dialog for linking projects. Should this also be transformed to DataTables (for consistancy)?

mcjaeger:
good points. Agree on the merge of buttons and the project linking table shall be transformed as well. If is not there there must be the version of a project displayed. And I think project state and project clearing state are useful infos as well.

ChristophNiehoff:
Converted also the project search to DataTables, added the missing columns and cleaned up the javascript.

ChristophNiehoff:
@mcjaeger ok, I will also change the table for linking releases to releases. But I would suggest to wait until issue #526 is resolved.

ChristophNiehoff:
Replaced the release search in the release edit page.

backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ProjectRepository.java
```diff
@@ -250,7 +250,7 @@ public ProjectRepository(DatabaseConnector db) {
     }
 
     public List<Project> searchByName(String name, User user) {
-        return searchByName(name, user, SummaryType.SHORT);
+        return searchByName(name, user, SummaryType.SUMMARY);
```
ChristophNiehoff: I build up a testing database with around 10.000 dummy projects and did not see any significant change in performance when switching from `SummaryType.SHORT` to `SummaryType.SUMMARY`. 

But with such a large number of projects the rendering of the DataTable (sorting) becomes time-consuming...

frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/tags/DisplayStateBoxes.java
```diff
@@ -0,0 +1,79 @@
...
+import javax.servlet.jsp.JspException;
+import javax.servlet.jsp.tagext.JspTag;
+import javax.servlet.jsp.tagext.SimpleTagSupport;
+import java.io.IOException;
+
+import static org.eclipse.sw360.datahandler.thrift.projects.ProjectClearingState.*;
+
+public class DisplayStateBoxes extends SimpleTagSupport {
+
+    public static final String PROJECT_STATE_ACTIVE__CSS      = "projectStateActive";
```
maxhbr: String also used in `frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/tags/DisplayProjectStateBox.java`. Please move constant to the next general place and use it in both places.

maxhbr: e.g. `frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/common/PortalConstants.java`

maxhbr:
@ChristophNiehoff (from discussion): please consider adding a processing-spinner, e.g. something like https://datatables.net/plug-ins/api/fnProcessingIndicator (which might not be on webjars)

ChristophNiehoff:
Added a "loading" and "processing"-indicator for the projects- and releases search tables.

mcjaeger:
tested the progress info from my side. we could merge it?

frontend/sw360-portlet/src/main/webapp/html/components/editRelease.jsp
```diff
@@ -241,9 +241,9 @@
             var beforeComma = vendorInfo.substr(0, vendorInfo.indexOf(","));
             var afterComma = vendorInfo.substr(vendorInfo.indexOf(",") + 1);
 
-            $('#<%=Release._Fields.VENDOR_ID.toString()%>').val(beforeComma.trim());
-            $('#<%=Release._Fields.VENDOR_ID.toString()%>Display').val(afterComma.trim());
-        }
+             $('#<%=Release._Fields.VENDOR_ID.toString()%>').val(beforeComma.trim());
```
alexbrdn: @ChristophNiehoff, I don't think you've added these spaces with a particular purpose in mind, have you? Could you, maybe, remove them real quick so that I coulc merge this PR?

frontend/sw360-portlet/src/main/webapp/html/projects/edit.jsp
```diff
@@ -29,7 +29,7 @@
 </portlet:actionURL>
 
 <portlet:actionURL var="deleteURL" name="delete">
-    <portlet:param name="<%=PortalConstants.PROJECT_ID%>" value="${project.id}"/>
+    <portlet:param name=    "<%=PortalConstants.PROJECT_ID%>" value="${project.id}"/>
```
alexbrdn: this whitespaec seems wasteful as well

ChristophNiehoff:
Oh yes, you are right.  I have changed that.

PR #536, opened 2017-08-16
fix(chore): fix outdated `liferay-plugin-package.properties` files

maxhbr:


maxhbr:
Note: In the file we define `module-group-id` (inconsistently) which is not mentioned in the corresponding specification: https://docs.liferay.com/portal/6.2/definitions/liferay-plugin-package_6_2_0.dtd.html. I think that this is an ignored field and can be removed.

Issue #436, opened 2017-06-08
Make external ids editable


mcjaeger:
Currently for external ids for releases, they cannot be edited by normal users (including creator).

Should be editable like flexible roles for project.

maxhbr:
assigned to @stevinus1

PR #537, opened 2017-08-18
feat(ui): Making external ids editable from release

stevinus1:
- Creates new editable table to represent external Ids in edit release view, which can be used to update the external ids.
screenshot from 2017-08-18 11-58-19
- Modifies the getCustomMapFromRequest method in PortletUtils so that it can be used for external id map.

Edit: Updating branch so it is up-to-date with master

Closes sw360/sw360portal#436

### DIFF
--- a/backend/src/src-moderation/src/main/java/org/eclipse/sw360/moderation/db/ModerationRequestGenerator.java
+++ b/backend/src/src-moderation/src/main/java/org/eclipse/sw360/moderation/db/ModerationRequestGenerator.java
@@ -26,6 +26,7 @@ import org.apache.thrift.protocol.TType;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Set;
+import java.util.Objects;
 import java.util.stream.Collectors;
 
 import static org.eclipse.sw360.datahandler.common.CommonUtils.add;
@@ -179,6 +180,32 @@ public abstract class ModerationRequestGenerator<U extends TFieldIdEnum, T exten
 
         documentAdditions.setFieldValue(field, addMap);
         documentDeletions.setFieldValue(field, deleteMap);
+    }
+
+    protected void dealWithStringtoStringMap(U field) {
+        Map<String, String> updateDocumentMap = CommonUtils.nullToEmptyMap(
+                (Map<String, String>) updateDocument.getFieldValue(field));
+        Map<String, String> actualDocumentMap = CommonUtils.nullToEmptyMap(
+                (Map<String, String>) actualDocument.getFieldValue(field));
+        if (updateDocumentMap.equals(actualDocumentMap)) {
+            return;
+        }
+
+        Map<String,String> addedMap = updateDocumentMap;
+        Map<String, String> deletedMap = actualDocumentMap;
+
+        Set<String> commonKeys = Sets.intersection(updateDocumentMap.keySet(), actualDocumentMap.keySet());
+        for(String id : commonKeys) {
+            String actual = actualDocumentMap.get(id);
+            String update = updateDocumentMap.get(id);
+            if(! Objects.equals(update, actual)) {
+                addedMap.put(id, update);
+                deletedMap.put(id, actual);
+            }
+        }
+
+        documentAdditions.setFieldValue(field, addedMap);
+        documentDeletions.setFieldValue(field, deletedMap);
     }
 
     protected void dealWithAttachments(U attachmentField){
--- a/frontend/sw360-portlet/src/main/webapp/html/components/editRelease.jsp
+++ b/frontend/sw360-portlet/src/main/webapp/html/components/editRelease.jsp
@@ -112,9 +112,11 @@
                             <core_rt:set var="mapTitle" value="Additional Roles"/>
                             <core_rt:set var="inputType" value="email"/>
                             <core_rt:set var="inputSubtitle" value="Enter mail address"/>
-
                             <core_rt:set var="customMap" value="${release.roles}"/>
                             <%@include file="/html/utils/includes/mapEdit.jspf" %>
+
+                            <core_rt:set var="externalIdsSet" value="${release.externalIds.entrySet()}"/>
+                            <%@include file="/html/components/includes/releases/editExternalIds.jsp" %>
                             <%@include file="/html/components/includes/releases/editReleaseRepository.jspf" %>
                         </div>
                         <div id="tab-ReleaseLinks">
@@ -193,6 +195,20 @@
 
 	        $('#formSubmit').click(
 	            function() {
+                    var valid = true;
+                    $.each($('#externalIdsTable tr.bodyRow'), function (i, row) {
+                        var key = $('#'+row.id+' td input.keyClass').val();
+                        var value = $('#'+row.id+' td input.valueClass').val();
+                        var isOneEmpty = ((key !== '') && (value === '')) || ((key === '') && (value !== ''));
+                        if (isOneEmpty) {
+                            $.alert('There is a blank field in one or more of your external ids. Please correct this before updating the release.');
+                            valid = false;
+                            return false;
+                        }
+                    });
+                    if (!valid) {
+                        return;
+                    }
 	                <core_rt:choose>
 	                    <core_rt:when test="${addMode || release.permissions[WRITE]}">
 	                        $('#releaseEditForm').submit();
--- a/frontend/sw360-portlet/src/main/webapp/html/components/includes/releases/editExternalIds.jsp
+++ b/frontend/sw360-portlet/src/main/webapp/html/components/includes/releases/editExternalIds.jsp
@@ -1,0 +1,59 @@
+<%@ page import="org.eclipse.sw360.portal.common.PortalConstants" %>
+<%@ page import="org.eclipse.sw360.datahandler.thrift.components.Release" %>
+<script src="<%=request.getContextPath()%>/webjars/jquery-validation/1.15.1/jquery.validate.min.js" type="text/javascript"></script>
+<script src="<%=request.getContextPath()%>/webjars/jquery-validation/1.15.1/additional-methods.min.js" type="text/javascript"></script>
+<table class="table info_table" id="externalIdsTable">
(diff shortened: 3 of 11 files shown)


frontend/sw360-portlet/src/main/webapp/html/components/editExternalIds.jsp
```diff
@@ -0,0 +1,63 @@
...
+            $('#' + rowIdOne).remove();
+        };
+
+        deleteConfirmed("Do you really want to remove this item?", deleteMapItemInternal);
+    }
+
+    function addRowToExternalIdsTable() {
+        var randomRowId = "mapItemRow" + Date.now();
+        var newRowAsString =
+            '<tr id="' + randomRowId + '" >'+
```
alexbrdn: Here, you basically repeat the same html code for a table row, but in a JS string. I don't like the repetition.
I see two options to get rid of it:
1. See editing of linked releases and projects. Look up usages of `linkedProjectsAjax.jsp` and go from there
2. See the new and modern way of achieving the same thing with JS and JSON in PR#530 by @lepokle: https://github.com/sw360/sw360portal/pull/530/files#diff-a2a329d1540568537c8fb67d0e245df3

stevinus1:
@alexbrdn 
I made the styling changes and also initialized the table using the addRowtoExternalIdsTable, so there is no code repetition. I hope this is OK?

frontend/sw360-portlet/src/main/webapp/html/components/includes/releases/editExternalIds.jsp
```diff
@@ -0,0 +1,59 @@
...
+
+        deleteConfirmed("Do you really want to remove this item?", deleteMapItemInternal);
+    }
+
+    function addRowToExternalIdsTable(key, value, rowId) {
+        if (!rowId) {
+            var rowId = "externalIdsTableRow" + Date.now();
+        }
+        if ((!key) || (!value)) {
+            var key = "", value = "";
```
alexbrdn: I don't know how bad that is, but if the key is null or undefined, this line will overwrite value with an empty string even if value was not empty, and vice versa.

stevinus1: Yeah, you're right. I changed it to && and also added a small function to prevent people from creating external ids with one blank field. I think this fixes the problem?

frontend/sw360-portlet/src/main/webapp/html/components/editRelease.jsp
```diff
@@ -244,6 +246,18 @@
 
         $('#formSubmit').click(
             function() {
+                var valid = true;
+                $.each($('#externalIdsTable tr.bodyRow'), function (i, row) {
+                    var key = $('#'+row.id+' td input.keyClass').val();
+                    var value = $('#'+row.id+' td input.valueClass').val();
+                    var isOneEmpty = ((key !== '') && (value === '')) || ((key === '') && (value !== ''));
```
alexbrdn: I don't mean to nitpick, but I still have a question.
`isOneEmpty` will be `true` if one and only one from both variables is empty. What happens if both are empty?

stevinus1: If both are empty nothing will be added to the database so there's no issue and the release can just update - @maxhbr suggested that this would make things easier if someone wants to create an external id and then just changes their mind.

alexbrdn: Ok, fine. I got it

frontend/sw360-portlet/src/main/webapp/html/components/editRelease.jsp
```diff
@@ -244,6 +246,18 @@
...
+                    if (isOneEmpty) {
+                        $.alert('There is a blank field in one of your external ids. Please correct this before updating the release.');
+                        valid = false;
+                        return false;
```
maxhbr: I think this return is unnecessary.

frontend/sw360-portlet/src/main/webapp/html/components/editRelease.jsp
```diff
@@ -244,6 +246,18 @@
...
+                    }
+                });
+                if (!valid) {return;}
```
maxhbr: please use three lines for this if statement.

frontend/sw360-portlet/src/main/webapp/html/components/editRelease.jsp
stevinus1: It prevents multiple alert windows appearing if there's more than one blank field

maxhbr: Ah, ok. I have overlooked the loop here.

libraries/lib-datahandler/src/main/java/org/eclipse/sw360/datahandler/common/Moderator.java
```diff
@@ -90,7 +90,11 @@ protected T updateBasicField(U field, FieldMetaData fieldMetaData, T document, T
                                     (Map<String, Set<String>>) documentAdditions.getFieldValue(field),
                                     (Map<String, Set<String>>) documentDeletions.getFieldValue(field)));
                 } else {
-                    log.error("Unknown field in Moderator: " + field.getFieldName());
+                    try {
+                        document.setFieldValue(field, documentAdditions.getFieldValue(field));
+                    } catch (ClassCastException e) {
```
stevinus1: @alexbrdn 
If you have time, I was wondering if you thought this bit of code is OK? 
I'm just adding in functionality so changes show up in moderation requests and I was unsure if using a try-catch block like this is good practice. There's an implicit cast in the call to documentAdditions.getFieldValue(field)

stevinus1:
@mcjaeger 
Right, I didn't think of that the first time.
I've added in the functionality now, I think the code is alright.

libraries/lib-datahandler/src/main/java/org/eclipse/sw360/datahandler/common/Moderator.java
alexbrdn: I see no problem with the try-catch block per se, but don't yet understand why it is necessary. Can you explain the use-case and the problem a bit more in detail?

stevinus1: Previously the code only updated the field after a moderation request is accepted if the field is a Map<String, Set<String>>. This edits it so it also updates if it's Map<String, String> - like the external Id field. I had an explicit cast in there so it said document.setFieldValue(field, (Map<String, String>)documentAdditions.getFieldValue(field)), but the IDE said it was redundant.

alexbrdn: I think I understand where you're coming from, but then I don't the try-catch block is necessary here at all. As I see it, just do `document.setFieldValue()` and you're good. If you want to add some special handling for the field `externalId` of `Release`, there's a good place in `ReleaseModerator` to do that. Otherwise, the one-line set should be enough, I think.

backend/src/src-moderation/src/main/java/org/eclipse/sw360/moderation/db/ModerationRequestGenerator.java
```diff
@@ -181,6 +181,32 @@ protected void dealWithCustomMap(U field) {
...
+        }
+
+        Map<String,String> addedMap = updateDocumentMap;
+        Map<String, String> deletedMap = actualDocumentMap;
+
+        Set<String> commonKeys = Sets.intersection(updateDocumentMap.keySet(), actualDocumentMap.keySet());
+        for(String id : commonKeys) {
+            String actual = actualDocumentMap.get(id);
+            String update = updateDocumentMap.get(id);
+            if(! actual.equals(update)) {
```
alexbrdn: That's a good candidate for a NullPointerException.

libraries/lib-datahandler/src/main/java/org/eclipse/sw360/datahandler/common/Moderator.java
stevinus1: OK, I get you, I've changed it. Thanks for your help.

backend/src/src-moderation/src/main/java/org/eclipse/sw360/moderation/db/ModerationRequestGenerator.java
```diff
@@ -181,6 +181,37 @@ protected void dealWithCustomMap(U field) {
...
+        Set<String> commonKeys = Sets.intersection(updateDocumentMap.keySet(), actualDocumentMap.keySet());
+        for(String id : commonKeys) {
+            String actual = actualDocumentMap.get(id);
+            String update = updateDocumentMap.get(id);
+            try {
+                if(! actual.equals(update)) {
+                    addedMap.put(id, update);
+                    deletedMap.put(id, actual);
+                }
+            } catch (NullPointerException e) {
```
alexbrdn: This is exactly the wrong solution. As a general rule, you don't ever catch a NullPointerException. You check the reference for null before dereferencing it.
In this specific case, comparing with `Objects.equals()` would be a sufficient and safe alternative, because it already handles null values correctly.

stevinus1: Hah, OK. Good to know for future :+1:

mcjaeger:
@stevinus1 shall I retest? (with the moderation requests)

stevinus1:
@mcjaeger Yes, thanks

mcjaeger:
@adrian-evo would it be Ok, if you test edit external ids with moderation requests?

mcjaeger:
@stevinus1 actually seems to work well, many thanks for that useful contribution!

I found https://github.com/sw360/sw360portal/issues/590 but maybe this is not related to this PR, so I decided to merge this.

Issue #459, opened 2017-06-13
Moderation (list) view shall inform about OSS vs. COTS type


mcjaeger:
The moderation request is sometimes for different users more or less important if it is about OSS or COTS. for example, there are COTS experts only interested in COTS related moderation requests.

As such, there should be an information about the component types for components and releases. This could be already in the list view, but should be at least in the moderation request detail view.

mcjaeger:
could be combined with https://github.com/sw360/sw360portal/issues/277

PR #529, opened 2017-08-09
feat(ui): Added ability to comment on moderation requests

ChristophNiehoff:
* Changes in the data structure:
	- Added comments of requesting user and moderator to moderation.thrift
	- Added user group of submitter to moderation.thrift (for display in the moderations details table)
	- Added `commentMadeDuringModerationRequest` to users.thrift
* Added ability for the moderator to give a comment on his/her decision
* Added ability for the submitter to give a comment on his/her request
* Cleaned up the moderation details table

Closes sw360/sw360portal#277

PR #538, opened 2017-08-21
feat(ui): Added component type to the listing of moderation requests

ChristophNiehoff:
- Added `ComponentType` to `moderation.thrift`
- For components and releases this is automatically set. For releases the type of the parent component is chosen. For other moderation requests (e.g. for projects) this field is not set.
- The moderation request listing now contains an additional column displaying the component type.

Closes sw360/sw360portal#459

### DIFF
--- a/libraries/lib-datahandler/src/main/thrift/moderation.thrift
+++ b/libraries/lib-datahandler/src/main/thrift/moderation.thrift
@@ -28,6 +28,7 @@ typedef projects.Project Project
 typedef users.User User
 typedef licenses.License License
 typedef licenses.Todo Todo
+typedef components.ComponentType ComponentType
 
 enum DocumentType {
     COMPONENT = 1,
@@ -56,9 +57,9 @@ struct ModerationRequest {
     17: optional string reviewer,
     18: required bool requestDocumentDelete,
     19: optional string requestingUserDepartment,
+    40: optional ComponentType componentType, // only relevant if the request is about components or releases
     34: optional string commentRequestingUser,
     35: optional string commentDecisionModerator,
-
     // Underlying objects
     20: optional Component componentAdditions,
     21: optional Release releaseAdditions,
--- a/scripts/migrations/008_add_component_type_to_moderation_requests.py
+++ b/scripts/migrations/008_add_component_type_to_moderation_requests.py
@@ -1,0 +1,110 @@
+#!/usr/bin/python
+# -----------------------------------------------------------------------------
+# Copyright Siemens AG, 2017. Part of the SW360 Portal Project.
+#
+# All rights reserved.   This program and the accompanying materials
+# are made available under the terms of the Eclipse Public License v1.0
+# which accompanies this distribution, and is available at
+# http://www.eclipse.org/legal/epl-v10.html
+#
+# This is a manual database migration script. It is assumed that a
+# dedicated framework for automatic migration will be written in the
+# future. When that happens, this script should be refactored to conform
+# to the framework's prerequisites to be run by the framework. For
+# example, server address and db name should be parametrized, the code
+# reorganized into a single class or function, etc.
+#
+# initial author: christoph.niehoff@tngtech.com
+#
+# -----------------------------------------------------------------------------
+import couchdb
+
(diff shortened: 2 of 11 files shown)


ChristophNiehoff:
After #529 is merged, the component type can also be added to the new moderation request detail view.

ChristophNiehoff:
Added a migration script to include the component type to already existing moderation requests in the database

libraries/lib-datahandler/src/main/thrift/moderation.thrift
```diff
@@ -55,6 +56,7 @@ struct ModerationRequest {
     17: optional string reviewer,
     18: required bool requestDocumentDelete,
     19: optional string requestingUserDepartment;
+    40: optional ComponentType componentType; // only relevant if the request is about components or releases
```
alexbrdn: I'd prefer to refrain from extending moderation requests with field that is specific for some document types. How about making sure that for component and release moderation requests, the component type is included in both `documentAdditions` and `documentDeletions` (thus indicating no change if not specifically requested) and using some utility method for extracting this from `documentAdditions` for displaying in the table? The utility method would default to empty for other moderation document types. Do you think that would work?

I think by going this route, you can even skip the db migration altogether. Moderation requests are relatively short-lived. I don't think it would be a big problem if you displayed the component type only for new moderation requests.

scripts/migrations/008_add_component_type_to_moderation_requests.py
```diff
@@ -0,0 +1,109 @@
...
+# This is a manual database migration script. It is assumed that a
+# dedicated framework for automatic migration will be written in the
+# future. When that happens, this script should be refactored to conform
+# to the framework's prerequisites to be run by the framework. For
+# example, server address and db name should be parametrized, the code
+# reorganized into a single class or function, etc.
+#
+# initial author: christoph.niehoff@tngtech.com
+#
+-----------------------------------------------------------------------------
```
alexbrdn: The hash sign has gone missing. This is not a comment now.

scripts/migrations/008_add_component_type_to_moderation_requests.py
```diff
@@ -0,0 +1,109 @@
...
+                componentID = document[COMPONENT_ID]
+                componentType = componentData[componentID]
+                updateDocument(db, document, {COMPONENT_TYPE: componentType})
+
+            elif (document[MODERATION_TYPE] == RELEASE) and (RELEASE_ID in document.keys()):
+                releaseID = document[RELEASE_ID]
+                componentType = releaseData[releaseID]
+                updateDocument(db, document, {COMPONENT_TYPE: componentType})
+
+print "done."
```
alexbrdn: no newline at eof

scripts/migrations/README.md
```diff
@@ -23,6 +23,10 @@ To migrate it is recommended to do this in the following order:
 ### 1.7.0 -> 2.0.0
 - `005_convert_compatibility_fields_to_ternary.py`
 - `006_convert_project_release_relationship_to_enums.py`
+### 2.0.0 -> 2.1.0
+### 2.1.0 -> 3.0.0
```
alexbrdn: we're at 2.2.0 now. please update this line

libraries/lib-datahandler/src/main/thrift/moderation.thrift
ChristophNiehoff: Yes, I agree that adding the `componentType` is not beautiful. But I am also not happy with your suggestion. First, `...Additions` and `...Deletions` should be about be about the proposed changes. I wouldn't like using them to transport some extra information. And second, I do not think this would work for releases as the `ComponentType` is not included in the `Release` class.

alexbrdn: You're right. This won't work. I don't think the usage of additions and deletions would be a problem, but the `Release` not having a `ComponentType` is definitely one.
Alright, then. Let's leave it as it is.

ChristophNiehoff:
Rebase on current master and corrected the typos in the python script

PR #540, opened 2017-08-22
feat(ui): select particular license for lincense info

lepokle:
This PR enables users to fine-tune the license info generation. They are now able not only
to select license files but also specific licenses in these files.


mcjaeger:
looks good, also no erformance issue with larger files,  but there seems to be an issue with some files:

2017-08-30 00_38_14-projects - sw360

could we have a session to debug?

backend/src/src-licenseinfo/src/test/java/org/eclipse/sw360/licenseinfo/parsers/SPDXParserTest.java
```diff
@@ -12,16 +12,13 @@
 package org.eclipse.sw360.licenseinfo.parsers;
 
 import org.eclipse.sw360.datahandler.common.SW360Constants;
-import org.eclipse.sw360.datahandler.common.UncheckedSW360Exception;
```
lepokle: does not exist anymore

libraries/lib-datahandler/src/main/thrift/licenseinfo.thrift
```diff
@@ -59,25 +61,23 @@ struct LicenseInfoParsingResult {
     32: optional string version,
 }
 
-service LicenseInfoService {
+struct LicenseInfoFile {
```
lepokle: This allows us to return more information with one call. This way, the backend must only be called once for license info generation.

lepokle:
@mcjaeger I'm not sure if we have checked the error from the above comment. Please check :-)

I've only one license file therefore my testing capabilites were limited. You should especially test one with acknowledgments.

frontend/sw360-portlet/src/main/webapp/html/projects/includes/attachmentSelectTable.jspf
```diff
@@ -122,25 +152,206 @@
...
+
+                function expandLicenseFile($linkedProjectsInfoTable, node) {
+                    var url,
+                        urlData = $linkedProjectsInfoTable.data(),
+                        data = node.row.data();
+
+                    if(data.rowType !== 'attachment' || data.loaded) {
+                        return;
+                    }
+                    data.loaded = true;
```
alexbrdn: Here, you set `data.loaded` to true before it is actually loaded. I'm not sure if this can create a problem.

frontend/sw360-portlet/src/main/webapp/html/projects/licenseInfo.jsp
```diff
@@ -1,5 +1,5 @@
 <%--
-  ~ Copyright Siemens AG, 2016. Part of the SW360 Portal Project.
+  ~ Copyright Siemens AG, 2017. Part of the SW360 Portal Project.
```
alexbrdn: please extend copyright dates instead of overwriting. e.g. '2016-2017' (applies to all files)

libraries/lib-datahandler/src/main/java/org/eclipse/sw360/datahandler/common/WrappedException.java
```diff
@@ -0,0 +1,117 @@
+package org.eclipse.sw360.datahandler.common;
+
+import org.apache.thrift.TException;
+import org.eclipse.sw360.datahandler.thrift.SW360Exception;
+
+/**
+ * This class should help catching checked exceptions especially in lambda
```
alexbrdn: Neat! Kudos for cleaning this up.

libraries/lib-datahandler/src/main/thrift/licenseinfo.thrift
```diff
@@ -59,25 +61,23 @@ struct LicenseInfoParsingResult {
...
-     * Returns the CLI contained in the attachment, if any
-     * */
-//    LicenseInfoParsingResult getLicenseInfoForAttachment(1: Attachment attachment);
+service LicenseInfoService {
 
     /**
-     * Returns the CLI for the given release.
-     * */
-//    LicenseInfoParsingResult getLicenseInfoForRelease(1: Release release, 2: string selectedAttachmentContentId);
+     * parses the attachment of one releases for license information and returns the result.
```
alexbrdn: should be "one release" (singular), probably

lepokle:
@alexbrdn thanks, comments processed

PR #542, opened 2017-08-23
feat(exports): add component type to release row in spreadsheet exports

imaykay:
* added component type to every row that ReleaseHelper creates
  * therefore has to batch load parent components of releases in more cases
  * has been part of additional data only for clearing status exports
  * now this additional data is only the project origin

closes sw360/sw360portal#326


imaykay:
The issue #326 to this PR contains a table in one of the last comments which might be helpful to understand what has been implemented...

Issue #543, opened 2017-08-25
Cannot edit projects on Firefox


adrian-evo:
On the latest 2.3.0-SNAPSHOT, when trying to edit a project, the Update Project is not working on Firefox. On the Chrome and IE it is working. Tried also with clean cache. The following is displayed in the console:
![edit](https://user-images.githubusercontent.com/19324942/29706796-b8419ee6-898b-11e7-923c-bffd0420b8c6.PNG)

imaykay:
Hi @adrian-evo - did you clean your cache in all three browsers or just firefox?

mcjaeger:
hi, I have seen this too. What I observed:
* error is after clean vagrant deployment
* error is in Firefox, Safari
* it is just the "Update Project" at both locations (upper left, lower left)
* even after complete deletion of browser data

PR #544, opened 2017-08-29
fix(projects): "Update Project" button is working again

imaykay:
* refactored javascript to be in require block so that jquery and its validation plugin are available when registering the form for validation

fixes sw360/sw360portal#543

First commit is the same as on https://github.com/sw360/sw360portal/pull/528 but I needed the code. So only the commit 9b7ecda has to be reviewed here.


imaykay:
@alexbrdn since the code changes are pretty similar to the ones in PR https://github.com/sw360/sw360portal/pull/528 it might be a good idea that you have a quick look on this issue as well.

Issue #446, opened 2017-06-09
Clearign status numbers for releases dos not contain sub projects


mcjaeger:
There can be projects with `0 0 0 0 0` which is in most cases just a container project that contains subprojects with releases. In this case it would be good to have the releases of the linked projects also in.

Also the consensus seem to change the view from the five numbers to `X of Y` like `23/50` meaning that 23 releases out of 50 are approved.

Adding to that: going over subprojects, the already async loading od the numbers will be super slow. Maybe there would be a way to load the numbers of the projects in batches (and not for all projects all at once)?

imaykay:
Okay, while implementing the feature, questions around permissions arose:

Let's assume there are
* three projects (P1, P2 and P3)

and 
* two Users (U1 and U2).

U1 has the permission to view P1 and P2
U2 has at least permission for P2 and P3
P1 contains P2
P2 contains P3

Shall the clearing summary of P3 be included in the projects list view of U1 for P1? As probably one of the upcoming feature requests after this story will be: show me which subproject is the one that is missing its clearings this question can be answered easily for me. In this case you will have to show U1 that there is a project P3 which he should not know of. But if do not aggregate P3's clearing summary, U1 might think that everything is cleared and go ahead.

So: respect permissions in clearing summary or not?

alexbrdn:
If I recall correctly, right now, all types of relations of projects to subprojects and projects to releases are followed in the counting of the five status numbers. I'd venture to say that this should remain so.

mcjaeger:
I think it should be the global view on the numbers of project listing regardless of the permissions

regarding relation: I would be OK not to change it or consider relation, but it would be better to do so. But I would count only 'referred' or 'duplicate' as the ignoring relations, because 'unknown' might be generally there when it is meant 'contained'

imaykay:
Alright, permissions are not checked and I ignore project relations "referred" and "duplicate" and release relations "referred".

PR #546, opened 2017-08-30
feat(projects): include subprojects' clearings in parent clearings in…

imaykay:
… project table

* added thrift service to fill clearing summary for project list including subproject data
  * wrote own project tree traversal as existing methods seemed too complex
  * permission check is only prepared, not performed
  * relation follow / no-follow is subject to change
* changed portlet to use new service that includes subproject data
* added paged requests in gui, currently with a batch size of 10 (meaning clearing summary is queried per ajax for 10 projects in one request)
  * requests are triggered sequentially, so they are non blocking but a request is only started after a former one finishes
  * to be able to query the first visible projects, the client sorting has been duplicated on the server
* changed the way clearing summary is displayed to a more condensed version like 2/7 meaning 2 of 7 releases are cleared

closes sw360/sw360portal#446

### DIFF
--- a/backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ProjectDatabaseHandler.java
+++ b/backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ProjectDatabaseHandler.java
@@ -25,7 +25,6 @@ import org.eclipse.sw360.datahandler.couchdb.AttachmentConnector;
 import org.eclipse.sw360.datahandler.couchdb.DatabaseConnector;
 import org.eclipse.sw360.datahandler.entitlement.ProjectModerator;
 import org.eclipse.sw360.datahandler.thrift.*;
-import org.eclipse.sw360.datahandler.thrift.components.Component;
 import org.eclipse.sw360.datahandler.thrift.components.Release;
 import org.eclipse.sw360.datahandler.thrift.components.ReleaseClearingStateSummary;
 import org.eclipse.sw360.datahandler.thrift.components.ReleaseLink;
@@ -41,16 +40,21 @@ import org.eclipse.sw360.mail.MailUtil;
 import org.ektorp.http.HttpClient;
 
 import java.net.MalformedURLException;
+import java.time.Instant;
 import java.util.*;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Function;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 
-import static org.eclipse.sw360.datahandler.common.CommonUtils.*;
+import static org.eclipse.sw360.datahandler.common.CommonUtils.isInProgressOrPending;
+import static org.eclipse.sw360.datahandler.common.CommonUtils.nullToEmptyList;
+import static org.eclipse.sw360.datahandler.common.CommonUtils.nullToEmptyMap;
 import static org.eclipse.sw360.datahandler.common.SW360Assert.assertNotNull;
 import static org.eclipse.sw360.datahandler.common.SW360Assert.fail;
-import static org.eclipse.sw360.datahandler.common.SW360Utils.*;
+import static org.eclipse.sw360.datahandler.common.SW360Utils.getBUFromOrganisation;
+import static org.eclipse.sw360.datahandler.common.SW360Utils.getCreatedOn;
+import static org.eclipse.sw360.datahandler.common.SW360Utils.printName;
 import static org.eclipse.sw360.datahandler.permissions.PermissionUtils.makePermission;
 
 /**
@@ -70,6 +74,23 @@ public class ProjectDatabaseHandler {
     private final AttachmentConnector attachmentConnector;
     private final ComponentDatabaseHandler componentDatabaseHandler;
     private final MailUtil mailUtil = new MailUtil();
+
+    // this caching structure is only used for filling clearing state summaries and
+    // should be avoided anywhere else so that no other outdated information will be
+    // provided
+    // for the clearing state it is
+    // 1. necessary because they take a long time to get calculated which might be a
+    // good reason for service clients to query not for all projects at once, but to
+    // divide the queries in subsets of project (and we do not want to load the
+    // whole project map for each of this subset queries)
+    // 2. okay because the data normally "only moves forward", meaning that already
+    // approved clearing reports will not be revoked. So worst case of caching is
+    // that outdated data is provided, not wrong one (as it would be okay to see a
+    // cleared project that is displayed as not yet cleared - but it won't be okay
+    // to see a uncleared project that is displayed as cleared but isn't anymore)
+    private static final java.time.Duration ALL_PROJECTS_ID_MAP_CACHE_LIFETIME = java.time.Duration.ofMinutes(2);
+    private Map<String, Project> cachedAllProjectsIdMap;
+    private Instant cachedAllProjectsIdMapLoadingInstant;
 
     public ProjectDatabaseHandler(Supplier<HttpClient> httpClient, String dbName, String attachmentDbName) throws MalformedURLException {
         this(httpClient, dbName, attachmentDbName, new ProjectModerator(), new ComponentDatabaseHandler(httpClient,dbName,attachmentDbName));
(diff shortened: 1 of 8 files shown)


frontend/sw360-portlet/src/main/webapp/html/projects/view.jsp
```diff
@@ -300,47 +300,81 @@
...
+                function d(v) { return v == undefined ? 0 : v; }
+
+                var resultElementAsString = "<span class=\"clearingstate\" title=\"Necessary data not found on server!\">Not available</span>",
+                    releaseCount,
+                    approvedCount;
+
+                if (clearing) {
+                    releaseCount = d(clearing.newRelease) + d(clearing.underClearing) + d(clearing.underClearingByProjectTeam) + d(clearing.reportAvailable) + d(clearing.approved);
+                    approvedCount = d(clearing.approved);
+                    resultElementAsString = "<span class=\"clearingstate\" title=\"" + approvedCount + (approvedCount === 1 ? " release" : " releases") + " out of " + releaseCount + (approvedCount === 1 ? " has" : " have") + " been approved (including subprojects)!\">" + approvedCount + "/" + releaseCount + "</span>";
```
alexbrdn: I don't think this sentence deserves an exclamation mark. Secondly, to lessen the confusion of our users (or discourage misconceptions, if you will), we shouldn't say that releases have been approved, but only that they have approved clearing reports.

backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ProjectDatabaseHandler.java
```diff
@@ -441,4 +445,72 @@ public RequestStatus updateProjectVulnerabilityRating(ProjectVulnerabilityRating
         }
         return projects;
     }
+
+    public List<Project> fillClearingStateSummaryIncludingSubprojects(List<Project> projects, User user) {
+        // load all projects in a single db request first for performance reasons
+        final Map<String, Project> allProjectsIdMap = ThriftUtils.getIdMap(repository.getAll());
```
alexbrdn: So, here's the big problem I see with this approach.
You've chopped the request for clearing status into many requests and for each of these requests, you load the complete projects repository. That means, to display the projects page for one user, we send N/10+1 requests for all of the projects to the DB instead of just two. Given that the number of project is only going to increase, I believe this approach has no future.
Remember how we recently had to limit the number of components loaded for display on the main page to improve performance? When we have to do this for projects, it will only make the +1 part faster. N/10 calls for clearing status will still be loading all projects. Same applies to the sticky filtering by Group, which is already being used.
I'm afraid that there is no acceptable way of loading deep project tree structures on the fly. I think we will have to compute this information in some background jobs and load it from the DB cache when necessary.
Case in point: my dev system contains 1099 projects. It took 2:25 min for the projects page to completely load, including clearing statuses.

Let's discuss.

alexbrdn: Correction: the N/10 part _is_ affected by filtering the initial list of projects as the N gets smaller. I don't think, however, that this invalidates my argument.

Then there is the option of decreasing the number of calls by increasing the denominator, i.e. loading clearing statuses for more projects per request. I tested with 100 projects per request: 1:22 min. That's faster, but not qualitatively better.

imaykay: Hi @alexbrdn - sure, maybe as a start after the call today? And if we want to research in more depth in person, let's do this in two weeks when we meet.

frontend/sw360-portlet/src/main/webapp/html/projects/view.jsp
```diff
@@ -300,47 +300,81 @@
...
-		        } else {
-		            releaseCounts = "Not available";
-		        }
-
-		        return "<span title=\"new release, under clearing, under clearing by the project clearing team, report available, approved\">" + releaseCounts + "</span>";
-		    }
+            function loadClearingStateSummaries() {
+                var tableData = projectsTable.data(),
+                    queryPageSize = 10,
+                    numberOfQueryPages = Math.floor(tableData.length / queryPageSize),
```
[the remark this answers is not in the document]
imaykay: done in new commit

frontend/sw360-portlet/src/main/webapp/html/projects/view.jsp
imaykay: done in new commit

backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ProjectDatabaseHandler.java
imaykay: As discussed, I used somewhat of a cache to only query all projects once in 2 minutes. It would be good if you could retest with your bigger set of test data.

backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ProjectDatabaseHandler.java
```diff
@@ -447,80 +468,138 @@ public RequestStatus updateProjectVulnerabilityRating(ProjectVulnerabilityRating
...
-                MailConstants.SUBJECT_FOR_NEW_PROJECT,
-                MailConstants.TEXT_FOR_NEW_PROJECT,
-                SW360Constants.NOTIFICATION_CLASS_PROJECT, Project._Fields.SECURITY_RESPONSIBLES.toString(),
-                project.getName(), project.getVersion());
-        mailUtil.sendMail(SW360Utils.unionValues(project.getRoles()), user,
-                MailConstants.SUBJECT_FOR_NEW_PROJECT,
-                MailConstants.TEXT_FOR_NEW_PROJECT,
-                SW360Constants.NOTIFICATION_CLASS_PROJECT, Project._Fields.ROLES.toString(),
-                project.getName(), project.getVersion());
+        mailUtil.sendMail(project.getProjectResponsible(), MailConstants.SUBJECT_FOR_NEW_PROJECT,
```
alexbrdn: did you not like my carefully manually formatted code so much, or is it just the auto-formatting you applied indiscriminately?

frontend/sw360-portlet/src/main/webapp/html/projects/view.jsp
```diff
@@ -313,47 +313,81 @@
...
+                    numberOfQueryPages = Math.ceil(tableData.length / queryPageSize),
+                    queryPagesPageIds = [],
+                    pageIds,
+                    cell,
+                    i, j, idx;
+
+                for (i = 1; i <= numberOfQueryPages; i++) {
+                    pageIds = [];
+                    for (j = 0; j < queryPageSize; j++) {
+                        idx = ((i-1) * queryPageSize) + j;
```
alexbrdn: this is really a minor thing, but you have the `i`-loop from 1 till `<= numberOfQueryPages` and the at only place you use the `i ` you decrement it by 1. why not doing it the conventional way from 0 to `< numberOfQueryPages` ?

backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ProjectDatabaseHandler.java
imaykay: Ooops, I just moved the methods to the end of the class after rebasing and that way autoformat thought I "edited" those lines :-/ Will fix it in a second.

frontend/sw360-portlet/src/main/webapp/html/projects/view.jsp
imaykay: Haha, I thought that was the idea on using Math.ceil so that the loop loops over the visible page numbers :D But now I got you: It's all about the variable value being correct according to its name.

alexbrdn:
I tried it again with 10 and 100 projects per request. There was not much difference. 1:23 min vs. 1:15 min. Which suggests that most of the time is spent on the server iterating the hierarchies and tuning the number of requests will not help much.

alexbrdn:
@imaykay, could you please resolve the conflicting file? (the last time, hopefully)

imaykay:
[x] done, should be mergeable :-)

Issue #553, opened 2017-09-05
Cannot add or edit components, releases, ...


alexbrdn:
Testing with current master (7f7d391), I cannot add or edit components, because of the following exception:
```
09:15:09,052 ERROR [http-bio-8080-exec-25][PortletServlet:116] javax.portlet.PortletException: org.apache.jasper.JasperException: An exception occurred processing JSP page /html/components/edit.jsp at line 131__128: </core_rt:if>_129: _130: <script>_131:     var permissions = {  'write':  ${permissions.makePermission(component, currentUser).isActionAllowed(RequestedAction.DELETE)},_132:                          'delete': ${permissions.makePermission(component, currentUser).isActionAllowed(RequestedAction.WRITE)}_133:     };_134: ___Stacktrace: [Sanitized]
javax.portlet.PortletException: org.apache.jasper.JasperException: An exception occurred processing JSP page /html/components/edit.jsp at line 131__128: </core_rt:if>_129: _130: <script>_131:     var permissions = {  'write':  ${permissions.makePermission(component, currentUser).isActionAllowed(RequestedAction.DELETE)},_132:                          'delete': ${permissions.makePermission(component, currentUser).isActionAllowed(RequestedAction.WRITE)}_133:     };_134: ___Stacktrace: [Sanitized]
	at com.liferay.portlet.PortletRequestDispatcherImpl.include(PortletRequestDispatcherImpl.java:117)
	at com.liferay.util.bridges.mvc.MVCPortlet.include(MVCPortlet.java:362)
	at com.liferay.util.bridges.mvc.MVCPortlet.include(MVCPortlet.java:378)
	at org.eclipse.sw360.portal.portlets.components.ComponentPortlet.doView(ComponentPortlet.java:443)
...
Caused by: java.lang.NullPointerException
	at org.eclipse.sw360.datahandler.permissions.DocumentPermissions.getStandardPermissions(DocumentPermissions.java:87)
	at org.eclipse.sw360.datahandler.permissions.ComponentPermissions.isActionAllowed(ComponentPermissions.java:55)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)

```

Looks like this [line](https://github.com/sw360/sw360portal/pull/529/files#diff-8700f57c399b868e7362066689bed832R131) from #529 has created the problem.

alexbrdn:
[...]

PR #561, opened 2017-09-06
fix(components,releases,projects,licenses): fix NullPointerException in edit pages

alexbrdn:
- Using enum values directly in EL expressions doesn't work and that caused the NPE. Converting the Enum to a JSP variable beforehand allows working around the problem.
- removed unnecessary JS dictionaries with permissions as all objects already get filled with current user's permissions before edit.

closes sw360/sw360portal#553


frontend/sw360-portlet/src/main/webapp/html/projects/edit.jsp
```diff
@@ -156,7 +155,7 @@
                           '<core_rt:if test="${not empty project.attachments}" ><li><sw360:out value="${project.attachmentsSize}"/> attachments</li></core_rt:if>'  +
                           '<core_rt:if test="${not empty project.linkedProjects or not empty project.releaseIdToUsage or not empty project.attachments}" ></ul></core_rt:if>' +
                           '</div>'+
-                          '<div ' + styleAsHiddenIfNeccessary(permissions.delete) + '><hr><label class=\'textlabel stackedLabel\'>Comment your changes</label><textarea id=\'moderationDeleteCommentField\' class=\'moderationCreationComment\' placeholder=\'Comment on request...\'></textarea></div>';
+                          '<div ' + styleAsHiddenIfNeccessary(${component.permissions[DELETE] == true}) + '><hr><label class=\'textlabel stackedLabel\'>Comment your changes</label><textarea id=\'moderationDeleteCommentField\' class=\'moderationCreationComment\' placeholder=\'Comment on request...\'></textarea></div>';
```
ChristophNiehoff: Shouldn't this be `${project.permissions[DELETE] == true}` at the project edit page?  I think there is a copy/paste-error in all the `styleAsHiddenIfNeccessary`-calls.
Do you really need the `== true` at this point?

ChristophNiehoff:
With this change we can also remove again the addition of `currentUser` to the request in all the edit pages.

alexbrdn:
Removed the `currentUser`

ChristophNiehoff:
@alexbrdn 
Maybe we have been to fast in approving this pull request, meaning I was too fast with the code review. During testing I realized that  the evaluation of writing permissions for licenses is not correct, such that also admins will be asked for a moderation request comment.  I think the reason for this lies in the fact that `${licenseDetail.permissions[WRITE]}` is used to evaluate the permissions. But in the `LicensePortlet.prepareDetailView` this `License`-object is not filled with a permissions-map. `moderationLicenseDetail` should rather be used to get the permissions for licenses.

alexbrdn:
Oh, well... I feel a little embarrassed for merging it too quickly. As this does not appear to be a showstopper, I'm going to make myself a cup of tea and fix it slowly and thoroughly.

PR #562, opened 2017-09-06
feat(projects): add sanity check on project update to prevent deletin…

alexbrdn:
…g all linked releases or all linked projects at once

- add new RequestStatus `FAILED_SANITY_CHECK` for thrift calls
- when a project update intends to delete all release links or all project links and there are more than 5 of them, cancel the update and return the status `FAILED_SANITY_CHECK`. This should prevent erroneous accidental deletion of all links, which was reported by users in the last weeks. 
- When update fails with status `FAILED_SANITY_CHECK` show corresponding error message to the user

### DIFF
--- a/frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/common/ErrorMessages.java
+++ b/frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/common/ErrorMessages.java
@@ -37,13 +37,14 @@ public class ErrorMessages {
     public static final String EXTERNAL_ID_CANNOT_BE_EMPTY = "External ID cannot be empty.";
     public static final String PASSWORD_CANNOT_BE_EMPTY = "Password cannot be empty.";
     public static final String PASSWORDS_DONT_MATCH = "Password do not match.";
-    public static final String COULD_NOT_CREATE_USER_MODERATION_REQUEST = "Could not create user moderation request";
+    public static final String COULD_NOT_CREATE_USER_MODERATION_REQUEST = "Could not create user moderation request.";
     public static final String EMAIL_ALREADY_EXISTS = "Email already exists.";
     public static final String FULL_NAME_ALREADY_EXISTS = "Full name already exists.";
     public static final String EXTERNAL_ID_ALREADY_EXISTS = "External id already exists.";
     public static final String DEFAULT_ERROR_MESSAGE = "Request could not be processed.";
     public static final String DOCUMENT_NOT_AVAILABLE = "The requested document is not available.";
     public static final String LICENSE_SHORTNAME_TAKEN = "License shortname is already taken.";
+    public static final String UPDATE_FAILED_SANITY_CHECK = "Document update has been rejected; cannot delete all linked releases or projects at once.";
 
     //this map is used in errorKeyToMessage.jspf to generate key-value pairs for the liferay-ui error tag
     public static final ImmutableList<String> allErrorMessages = ImmutableList.<String>builder()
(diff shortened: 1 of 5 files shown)


frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/portlets/Sw360Portlet.java
```diff
@@ -263,6 +263,8 @@ public void setSessionMessage(PortletRequest request, RequestStatus requestStatu
                     setSW360SessionError(request, ErrorMessages.DOCUMENT_USED_BY_PROJECT_OR_RELEASE);
                 }
                 break;
+            case FAILED_SANITY_CHECK:
+                setSW360SessionError(request, ErrorMessages.UPDATE_FAILED_SANITY_CHECK);
```
ChristophNiehoff: A `break;` is missing at this place causing the following `PortletException` to be thrown.

frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/common/ErrorMessages.java
```diff
@@ -37,13 +37,14 @@
...
     public static final String PASSWORDS_DONT_MATCH = "Password do not match.";
-    public static final String COULD_NOT_CREATE_USER_MODERATION_REQUEST = "Could not create user moderation request";
+    public static final String COULD_NOT_CREATE_USER_MODERATION_REQUEST = "Could not create user moderation request.";
     public static final String EMAIL_ALREADY_EXISTS = "Email already exists.";
     public static final String FULL_NAME_ALREADY_EXISTS = "Full name already exists.";
     public static final String EXTERNAL_ID_ALREADY_EXISTS = "External id already exists.";
     public static final String DEFAULT_ERROR_MESSAGE = "Request could not be processed.";
     public static final String DOCUMENT_NOT_AVAILABLE = "The requested document is not available.";
     public static final String LICENSE_SHORTNAME_TAKEN = "License shortname is already taken.";
+    public static final String UPDATE_FAILED_SANITY_CHECK = "Document update has not been performed, as it failed the sanity check.";
```
ChristophNiehoff: I think the normal user does not know what "the sanity check" is. We should tell the user explicitly that he/she cannot delete ALL linked releases or projects at once, if the there are more than 5 of them.

mcjaeger: yes, maybe mention that "too many items (>5) were attemped to delete at once".

mcjaeger:
hm, need to go over this together:

* testing with admin
1. logging in
2. open project with access rights (=same group)
3. add 7 new releases
4. save
5. edit again
6. removing 7 releases
-> nothing new, just accepting the change

* test with clearing admin but not same group
1. opening project from different group
2. edit
3. remove all 6 entries, so the project has no releases anymore from its original 6
4. enter some moderation request reason
5. save
-> error "Proejcts not available"

on Version: 3.1.0-SNAPSHOT | Branch: feat/sanityChecksOnUpdate (4a4c3b6) | Build time: 20170911-0935

with stack trace:
 10:00:46,806 ERROR [http-nio-8080-exec-212][PortletServlet:116] javax.portlet.PortletException: javax.portlet.PortletException: Unknown request status
javax.portlet.PortletException: javax.portlet.PortletException: Unknown request status
-> i will retest, maybe a deployment problem

ChristophNiehoff:
@mcjaeger I think the last error `javax.portlet.PortletException: Unknown request status` is due to the missing `break` as mentioned in my code review.

mcjaeger:
after 
* cleam maven repo
* clean container
-> error is still there: "Portlet is temporarily unavailable. " after submitting moderation request message

alexbrdn:
I have definitely messed up something between testing for myself and committing. Sorry about that. Let me fix it

mcjaeger:
I rested it, works, also moderation request works, because no moderation request is created. :+1:

Issue #457, opened 2017-06-12
Improve approval of attachments and prevent deletion of approved attachments


mcjaeger:
A number of things came across the handling with attachments:

* If an attachment is approved it should be deleted anymore.
* A warning should be displayed that the deletion is not successful and tell user that approval needs to be on rejected or not checked.
* In order to delete such attachment, the approval must be revoked (should be done by the user)
* Only clearing or other admins shall be able to approve attachments other will do a moderation request.

PR #572, opened 2017-09-08
feat(attachments): Prevent deletion of approved attachments

ChristophNiehoff:
* Attachments can only be deleted if they are not approved (enforced this in the backend)
* In the GUI a warning is displayed if the user wants to delete an approved attachment, that does not allow him to do so.

Closes sw360/sw360portal#457

### DIFF
--- a/backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/AttachmentAwareDatabaseHandler.java
+++ b/backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/AttachmentAwareDatabaseHandler.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright Siemens AG, 2017. Part of the SW360 Portal Project.
+ *
+ * SPDX-License-Identifier: EPL-1.0
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+
+package org.eclipse.sw360.datahandler.db;
+
+import org.eclipse.sw360.datahandler.thrift.attachments.Attachment;
+import org.eclipse.sw360.datahandler.thrift.attachments.CheckStatus;
+
+import java.util.HashSet;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import static org.eclipse.sw360.datahandler.common.CommonUtils.nullToEmptyCollection;
+import static org.eclipse.sw360.datahandler.common.CommonUtils.nullToEmptySet;
+
+public abstract class AttachmentAwareDatabaseHandler {
+
+    public Set<Attachment> getAllAttachmentsToKeep(Set<Attachment> originalAttachments, Set<Attachment> changedAttachments) {
+        Set <Attachment> attachmentsToKeep = new HashSet<>();
+        attachmentsToKeep.addAll(nullToEmptySet(changedAttachments));
+        Set<String> alreadyPresentIdsInAttachmentsToKeep = nullToEmptyCollection(attachmentsToKeep).stream().map(Attachment::getAttachmentContentId).collect(Collectors.toSet());
+
+        // prevent deletion of already accepted attachments
+        Set<Attachment> acceptedAttachmentsNotYetToKeep = nullToEmptySet(originalAttachments).stream().filter(a -> (a.getCheckStatus() == CheckStatus.ACCEPTED && !alreadyPresentIdsInAttachmentsToKeep.contains(a.getAttachmentContentId()))).collect(Collectors.toSet());
+        attachmentsToKeep.addAll(acceptedAttachmentsNotYetToKeep);
+
+        return attachmentsToKeep;
+    }
+
+
+    public AttachmentAwareDatabaseHandler() {
+
+    }
+}
--- a/backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ComponentDatabaseHandler.java
+++ b/backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ComponentDatabaseHandler.java
@@ -69,7 +69,7 @@ import static org.eclipse.sw360.datahandler.thrift.ThriftValidate.prepareRelease
  * @author Johannes.Najjar@tngtech.com
  * @author alex.borodin@evosoft.com
  */
-public class ComponentDatabaseHandler {
+public class ComponentDatabaseHandler extends AttachmentAwareDatabaseHandler {
 
     private static final Logger log = Logger.getLogger(ComponentDatabaseHandler.class);
 
@@ -403,8 +403,10 @@ public class ComponentDatabaseHandler {
             component.unsetReleases();
 
             copyFields(actual, component, immutableOfComponent());
+            component.setAttachments( getAllAttachmentsToKeep(actual.getAttachments(), component.getAttachments()) );
             // Update the database with the component
             componentRepository.update(component);
+
             //clean up attachments in database
             attachmentConnector.deleteAttachmentDifference(actual.getAttachments(),component.getAttachments());
             sendMailNotificationsForComponentUpdate(component, user.getEmail());
@@ -463,7 +465,7 @@ public class ComponentDatabaseHandler {
             if (hasChangesInEccFields) {
                 autosetEccUpdaterInfo(release, user);
             }
-
+            release.setAttachments( getAllAttachmentsToKeep(actual.getAttachments(), release.getAttachments()) );
             releaseRepository.update(release);
             updateReleaseDependentFieldsForComponentId(release.getComponentId());
             //clean up attachments in database
@@ -584,6 +586,7 @@ public class ComponentDatabaseHandler {
     ///////////////////////////////
     // DELETE INDIVIDUAL OBJECTS //
     ///////////////////////////////
+
 
     public RequestStatus deleteComponent(String id, User user) throws SW360Exception {
         Component component = componentRepository.get(id);
--- a/backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ProjectDatabaseHandler.java
+++ b/backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ProjectDatabaseHandler.java
@@ -54,7 +54,7 @@ import static org.eclipse.sw360.datahandler.permissions.PermissionUtils.makePerm
  * @author daniele.fognini@tngtech.com
  * @author alex.borodin@evosoft.com
  */
-public class ProjectDatabaseHandler {
+public class ProjectDatabaseHandler extends AttachmentAwareDatabaseHandler {
 
     private static final Logger log = Logger.getLogger(ProjectDatabaseHandler.class);
     private static final int DELETION_SANITY_CHECK_THRESHOLD = 5;
@@ -182,6 +182,7 @@ public class ProjectDatabaseHandler {
             return RequestStatus.FAILED_SANITY_CHECK;
         } else if (makePermission(actual, user).isActionAllowed(RequestedAction.WRITE)) {
             copyImmutableFields(project,actual);
+            project.setAttachments( getAllAttachmentsToKeep(actual.getAttachments(), project.getAttachments()) );
             repository.update(project);
 
             //clean up attachments in database
--- a/frontend/sw360-portlet/src/main/webapp/html/utils/includes/editAttachments.jspf
+++ b/frontend/sw360-portlet/src/main/webapp/html/utils/includes/editAttachments.jspf
@@ -15,16 +15,18 @@
     Required Imports:       - org.eclipse.sw360.portal.common.PortalConstants
                             - org.eclipse.sw360.datahandler.thrift.attachments.Attachment
                             - org.eclipse.sw360.datahandler.thrift.components.Release
+                            - org.eclipse.sw360.datahandler.thrift.attachments.CheckStatus
 
     Required Beans:         - documentID
                             - documentType
+
+    Required variables:     - hasWritePermissions
 
     Required Stylesheets:   - jquery-ui
                             - jquery-confirm
 
     Included JSPFs:         - utils/includes/attachmentsUpload
 --%>
-
 <portlet:resourceURL var="attachmentListAjaxURL">
     <portlet:param name="<%=PortalConstants.ACTION%>" value='<%=PortalConstants.ATTACHMENT_LIST%>'/>
     <portlet:param name="<%=PortalConstants.DOCUMENT_ID%>" value="${documentID}"/>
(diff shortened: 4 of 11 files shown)


frontend/sw360-portlet/src/main/webapp/html/utils/includes/editAttachments.jspf
```diff
@@ -15,16 +15,19 @@
...
 
+    Required variables:     - hasWritePermissions
+
     Required Stylesheets:   - jquery-ui
                             - jquery-confirm
 
     Included JSPFs:         - utils/includes/attachmentsUpload
 --%>
-
+<%@ taglib prefix="core_t" uri="http://java.sun.com/jsp/jstl/core" %>
```
alexbrdn: `core_rt` is used as the prefix for JSTL core taglib throughout the codebase. Please adjust for consistency.

frontend/sw360-portlet/src/main/webapp/html/utils/includes/editAttachments.jspf
```diff
@@ -173,9 +179,15 @@
...
             var data = $(event.currentTarget).data();
-            deleteAttachment(data.key);
+            if(data.approvalState != <%=CheckStatus.ACCEPTED.getValue()%>) {
+                deleteAttachment(data.key);
+            } else {
+                displayDeleteNotAllowedWarning();
+            }
         });
 
+        $('#attachmentInfo ').on('change', function(event){  } )
```
alexbrdn: What's the point? Also, there's a trailing space in the selector.

frontend/sw360-portlet/src/main/webapp/html/utils/includes/editAttachments.jspf
```diff
@@ -198,6 +210,24 @@
 
             confirm.confirmDeletion("Do you really want to delete attachment <b>" + attachmentName + "</b> (" + attachmentId + ")?", deleteAttachmentInternal);
         }
+
+        function displayDeleteNotAllowedWarning() {
+            var warningMessage = "<p>An attachment cannot be deleted while it is approved. You have to reject the approval first.</p>";
+            <core_t:if test="${hasWritePermissions == false}">
+                warningMessage += "<p>For this you have to create a moderation request.</p>";
```
alexbrdn: I believe there must be a comma after "For this"

frontend/sw360-portlet/src/main/webapp/js/modules/datatables-renderer.js
```diff
@@ -237,10 +237,11 @@ define('modules/datatables-renderer', ['jquery', /* jquery-plugins */ 'datatable
...
      * <ol>
-     *   <li>key:   Will be set as "data-key" attribute to be able to identify the row</li>
-     *   <li>class: class name or space space separated class names for the action element</li>
-     *   <li>icon:  name of the icon to be shown, e.g. "Trash"
-     *   <li>title: Title to be shown on mouse over or if the icon could not be found</li>
+     *   <li>key:         Will be set as "data-key" attribute to be able to identify the row</li>
+     *   <li>class:       class name or space space separated class names for the action element</li>
+     *   <li>icon:        name of the icon to be shown, e.g. "Trash"
+     *   <li>title:       Title to be shown on mouse over or if the icon could not be found</li>
+     *   <li>approvalKey: Will be set as "data-approval-state" attribute. This allows to decide whether an attachment is deletable.</li>
```
alexbrdn: Maybe adjust the usage example at line 251 with this new property

libraries/lib-datahandler/src/main/java/org/eclipse/sw360/datahandler/couchdb/AttachmentConnector.java
```diff
@@ -75,26 +77,37 @@ public void deleteAttachment(String id) {
...
 
-    private Set<String> getAttachmentContenIds(Collection<Attachment> attachments) {
+    private Set<String> getAttachmentContentIds(Collection<Attachment> attachments) {
         return nullToEmptyCollection(attachments).stream()
                 .map(Attachment::getAttachmentContentId)
                 .collect(Collectors.toSet());
     }
 
-    public void deleteAttachmentDifference(Set<Attachment> before, Set<Attachment> after){
+    public void deleteAttachmentDifference(Set<Attachment> attachmentBefore, Set<Attachment> attachmentAfter){
```
alexbrdn: I don't mind the renaming, but then it should be plural, because these are sets. E.g. `attachmentsBefore`, `attachmentsAfter`.

libraries/lib-datahandler/src/main/java/org/eclipse/sw360/datahandler/couchdb/AttachmentConnector.java
```diff
@@ -75,26 +77,37 @@ public void deleteAttachment(String id) {
...
-        // otherwise, when `after` contains the same attachment (with the same id), but with one field changed (e.g. sha1),
+        // otherwise, when `attachmentAfter` contains the same attachment (with the same id), but with one field changed (e.g. sha1),
         // then they are considered unequal and the set difference will contain this attachment and therefore
-        // deleteAttachments(Collection<Attachment>) will delete an attachment that is present in `after`
-        deleteAttachmentsByIds(Sets.difference(getAttachmentContenIds(before), getAttachmentContenIds(after)));
+        // deleteAttachments(Collection<Attachment>) will delete an attachment that is present in `attachmentAfter`
+        Set<String> attachmentIdsAfter = getAttachmentContentIds(attachmentAfter);
+        Set<String> attachmentIdsToBeDeleted = new HashSet<>();
+
+        if(attachmentBefore != null) { // if the attachment prior to updating is null then it is either an invalid entry in the database or a test. In both case do nothing.
```
alexbrdn: That's a bunch of imperative code where it's not strictly necessary.
I think this code is neater and it passes the test.
```
        Set<Attachment> nonAcceptedAttachmentsBefore = nullToEmptySet(attachmentBefore).stream().filter(a -> a.getCheckStatus() != CheckStatus.ACCEPTED).collect(Collectors.toSet());
        deleteAttachmentsByIds(Sets.difference(getAttachmentContentIds(nonAcceptedAttachmentsBefore), getAttachmentContentIds(attachmentAfter)));

```

ChristophNiehoff:
Implemented the changes requested by @alexbrdn and rebased to current master.
I also realized that the previous implementation had an error: The backend prevented accepted attachments from being deleted. But the linking of attachments to projects/components/releases was still deletable, such that the attachments were not visible anymore in the GUI. Corrected this. @alexbrdn please have a re-review.

@adrian-evo What do you mean with "Added attachment as user"?  If the user also created the project, then he has write permissions anyway...  Changed the behavior such that also admins have to revoke their approval by updating the project. This is more consistent in terms of accepting/revoking moderation requests.

backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/AttachmentAwareDatabaseHandler.java
```diff
@@ -0,0 +1,38 @@
...
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import static org.eclipse.sw360.datahandler.common.CommonUtils.nullToEmptyCollection;
+import static org.eclipse.sw360.datahandler.common.CommonUtils.nullToEmptySet;
+
+public abstract class AttachmentAwareDatabaseHandler {
+
+    public void preventDeletionOfAcceptedAttachments(Set<Attachment> originalAttachments, Set<Attachment> changedAttachments) {
+        if(changedAttachments != null) {
```
alexbrdn: What happens if `changedAttachments` is null? It will allow deleting all attachments, including accepted ones? No bueno.
I think this method should return a `Set` that has to be used instead of `changedAttachments` by the caller. Then, the only thing you need to do here, is to get all accepted attachments from `originalAttachments` and add them to `changedAttachments`, then return that.

backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ComponentDatabaseHandler.java
```diff
@@ -463,7 +465,7 @@ public RequestStatus updateRelease(Release release, User user, Iterable<Release.
             if (hasChangesInEccFields) {
                 autosetEccUpdaterInfo(release, user);
             }
-
+            preventDeletionOfAcceptedAttachments(actual.attachments, release.attachments);
```
alexbrdn: Why don't you use the getters? IMO, that's preferable to direct field access.

backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ProjectDatabaseHandler.java
```diff
@@ -192,6 +190,7 @@ public RequestStatus updateProject(Project project, User user) throws SW360Excep
             return RequestStatus.FAILED_SANITY_CHECK;
         } else if (makePermission(actual, user).isActionAllowed(RequestedAction.WRITE)) {
             copyImmutableFields(project,actual);
+            preventDeletionOfAcceptedAttachments(actual.attachments, project.attachments);
```
alexbrdn: getters here, too?

backend/src-common/src/test/java/org/eclipse/sw360/datahandler/db/ProjectDatabaseHandlerTest.java
```diff
@@ -67,7 +63,7 @@ public void setUp() throws Exception {
 
         List<Project> projects = new ArrayList<>();
 
-        Project p1 = new Project().setId("P1").setName("Project1").setBusinessUnit("AB CD EF").setCreatedBy("user1")
+        Project p1 = new Project().setId("P1").setName("Project1").setBusinessUnit("AB CD EF").setCreatedBy("user1").setAttachments(Collections.emptySet())
```
alexbrdn: this is cheating ;) I see no reason why it shouldn't work when project's attachments are null. Or is there one?
Also, you can't be sure that there aren't projects in the DB with null attachments.
I suggest instead to add an extra test to make sure that null attachments won't cause problems and won't delete accepted attachments, either.

This applies to line 76 and changes in ProjectHandlerTest, as well.

backend/src-common/src/test/java/org/eclipse/sw360/datahandler/db/ProjectDatabaseHandlerTest.java
```diff
@@ -279,4 +275,6 @@ public void testSanityCheckSucceeds() throws Exception {
         RequestStatus status2 = handler.updateProject(project2, user1);
         assertThat(status2, is(RequestStatus.SUCCESS));
     }
+
```
alexbrdn: whitespace not strictly necessary here

adrian-evo:
"Added attachment as user" I mean normal user was able to remove the attachment without revoking the approval, but admin had to revoke the approval before deleting.

ChristophNiehoff:
Implemented the changes requested by @alexbrdn  and rebased to current master. Please have a re-review.

mcjaeger:
@adrian-evo is the proposed two-stage solution (first "de-accept", then second edit allows to remove) OK for you?

adrian-evo:
The two-stage flow is fine and working for when an user is trying to edit an attachment with moderation request. 

However, in my case I have the following issue:
- Login as Admin and edit a project created by User
- Add an attachment and approve it
- Delete the attachment. The message is allowing to delete it, however the attachment is not really deleted after Confirm and refresh the page. See attached.
attachment admin

ChristophNiehoff:
Updated to current master and removed the unused `import`. 

@adrian-evo  I am sorry, but following your steps I cannot reproduce the issue. Could you retry it with the current version?

adrian-evo:
Steps:
- Created a new project as user@sw360.org
- Add an attachment to this project, then change status to Accepted, update project
- Edit the project, remove the attachment.
Actual:
- The message "Do you really want to delete..." appear. Confirm.
- The attachment seems to be deleted, but then it is back on the list after refresh.

Do the same steps with admin@sw360.org
Actual:
- The message "An attachment cannot be deleted..." appear

user attachment
admin attachment

ChristophNiehoff:
@adrian-evo  Now I see the problem. It seems to be working fine on Chrome but there is a problem with Firefox.  I'll have a look at it.

ChristophNiehoff:
It seems that Firefox has a problem if HTML5 data()-attributes contain a second dash. Changed this. 
@adrian-evo  Thank you for finding this bug!  Could you retest if it works now?

adrian-evo:
The change seems to work as expected, however there is another issue now:
- Edit a project
- Change the Approval Status (e.g. from Accepted to Rejected)
- Update the project
- Click on Attachments

The list of attachments is empty although there is a header showing that there are attachments. The same if you click on another project with attachments.

project attach

ChristophNiehoff:
@adrian-evo I am sorry, but again I cannot reproduce the issue. Could you give me more information:
* In which browser are you seeing this?
* Are there error messages showing up in the console of the browser?  (From your screenshot it appears that the datatable did not initialize properly, so there should be javascript errors...)
* Does this really affect *all* projects with attachments?
* Can you see whether this is connected to issue https://github.com/sw360/sw360portal/issues/594 ?  I.e. does it work properly in Chrome?

adrian-evo:
It seems only in Firefox the list of attachment is empty, in Chrome it is not, however the error in console is displayed on both browsers. Tried with many projects with attachments and all are affected.
Attached the console from Chrome.

attach chrome

ChristophNiehoff:
@adrian-evo Could it be that all affected projects have linked projects or releases? It think we know this error from the "Linked projects and releases" table.   If that table throws a Javascript error, then the attachment table might not be rendered correctly.

adrian-evo:
It seems the issue reproduces also when creating a new project with no links, and adding an attachment to it. I can only get the attachment listed if I open a component, refresh browser and then go back to project.

mcjaeger:
Maybe this branch would take advatange from a rebase, because there seem to be Jva script issues as well:

*tested: feat/457-approval-of-attachments (173ba3f)
* In Edit view of  a release no attachments are visible at all (I am sure there are, but with branch no attachments are visible)
* In edit view of a release: nothing happen when clicking "Add Attachments"

some of the java script error message:
```
Error: Mismatched anonymous define() module: function (a){!function(){function b(a){return a.replace(/<.[^<>]*?>/g," ").replace(/&nbsp;|&#160;/gi," ").replace(/[.(),;:!?%#$'\"_+=\/\-“”’]*/g,"")}a.validator.addMethod("maxWords",function(a,c,d){return this.optional(c)||b(a).match(/\b\w+\b/g).length<=d},a.validator.format("Please enter {0} words or less.")),a.validator.addMethod("minWords",function(a,c,d){return this.optional(c)||b(a).match(/\b\w+\b/g).length>=d},a.validator.format("Please enter at least {0} words.")),a.validator.addMethod("rangeWords",function(a,c,d){var e=b(a),f=/\b\w+\b/g;return this.optional(c)||e.match(f).length>=d[0]&&e.match(f).length<=d[1]},a.validator.format("Please enter between {0} and {1} words."))}(),a.validator.addMethod("accept",function(b,c,d){var e,f,g,h="string"==typeof d?d.replace(/\s/g,""):"image/*",i=this.optional(c);if(i)return i;if("file"===a(c).attr("type")&&(h=h.replace(/[\-\[\]\/\{\}\(\)\+\?\.\\\^\$\|]/g,"\\$&").replace(/,/g,"|").replace(/\/\*/g,
```

ChristophNiehoff:
Rebased to the current master.

After the rebase there still was an error in the release edit page: When being in `addMode` (and only in this case) the addition of attachments did not work properly. 
For projects and components, attachments cannot be added immediately when creating a new project/component. One has to add the new object first and only after that attachments can be added. For releases this was not the case and I think PR https://github.com/sw360/sw360portal/pull/530 did not support this special case for releases. So, I made the edit page for release consistent to the ones of projects and releases: Attachments can only be added after the new release was created.
@mcjaeger Is this fine for you?

adrian-evo:
Works as expected now. Also I think it makes sense also for releases to be able to add attachments only after creating the object, which would be consistent with projects and components.

Issue #585, opened 2017-09-19
Moderation request not sent to admins anymore


adrian-evo:
Seems that lately the admins are not getting the moderation request, unless they are the creator of the item. E.g. if a project is created by Test Admin then the moderation request is sent to this admin, but if a project is created by Test User then the moderation request is sent only to this user but not admins.

Issue #582, opened 2017-09-14
License moderation broken


adrian-evo:
When adding a todo or edit white list as user, there is a moderation request message, however none of the admins get the moderation request.

ChristophNiehoff:
It appears that this is an unexpected side effect of pull request #551 . I will have a look at it.

ChristophNiehoff:
The problem seems to be the following:  During the creation of the moderation request `ModerationDatabaseHandler.getAllSW360Users()` is called. This calls in the end `UserSearch.saerchByNameAndEmail(null)`. The `null`-arguments clashes with the newly implemented lucene search.

PR #587, opened 2017-09-19
fix(moderation): fix bug in license moderation

ChristophNiehoff:
Closes sw360/sw360portal#582
Closes sw360/sw360portal#585


ChristophNiehoff:
@adrian-evo Could you test if this fixes https://github.com/sw360/sw360portal/issues/582 ?

adrian-evo:
Tested and works as expected now.

adrian-evo:
Also #585 is fixed with this update.

PR #589, opened 2017-09-20
fix(migration): Make migration script `008_add_component_type_to_mode…

ChristophNiehoff:
…ration_requests.py` more robust

- changed the migration script such that missing a `componentType` in a component does not result in an error during migration

### DIFF
--- a/scripts/migrations/008_add_component_type_to_moderation_requests.py
+++ b/scripts/migrations/008_add_component_type_to_moderation_requests.py
@@ -87,7 +87,8 @@ releaseData = {}
 for release in allReleases:
     document = release.value
     if PARENT_COMPONENT_ID in document.keys():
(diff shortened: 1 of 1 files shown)


scripts/migrations/008_add_component_type_to_moderation_requests.py
```diff
@@ -87,7 +87,8 @@ def updateDocument(db, doc, newDict):
 for release in allReleases:
     document = release.value
     if PARENT_COMPONENT_ID in document.keys():
-        releaseData[release.id] = componentData[document[PARENT_COMPONENT_ID]]
+        if document[PARENT_COMPONENT_ID] in componentData.keys():
```
alexbrdn: I would join these two `if`s.
Also, I believe it is Python convention to write `if key in dict`, as opposed to `if key in dict.keys()`

Issue #516, opened 2017-07-25
Search Vendor dialog should have aligned headers


adrian-evo:
Currently the dialog is a little bit odd, since e.g. short names and urls are far away left from the header names.
![vendor](https://user-images.githubusercontent.com/19324942/28569807-1c59fc94-7144-11e7-912c-baf999fd465d.PNG)

PR #600, opened 2017-09-28
fix(vendorsearch): align table headers

lepokle:
Use colgroups to make columns of both tables the same width.

It is nearly impossible to use only one table and make the tbody scrollable.

Closes sw360/sw360portal#516


mcjaeger:
@adrian-evo if you would have time, mind a test?

adrian-evo:
Works as expected.

Issue #394, opened 2017-05-10
Add Help Tips for Enums and Fields


alexbrdn:
(as discussed in our weekly call)

Some enum values that are used in object properties are not self-explanatory. For example, it is not at all obvious what some values of MainlineState (OPEN, SPECIFIC) mean.

To remedy some confunsion, it is proposed to create a custom tag that will take an [sw360 thrift] enum class as a parameter and create an `(i)` icon with a tooltip explaining the meaning of different values. This tag should then be inserted at various places where the values of enum types are displayed or edited.

It would probably be wise to somehow only render the help text once per page and call show it via javascript, otherwise a long table may include thousands of copies of the same help tooltip.

Addition from Michael: Could be actually more sort of help tool tips for fields to explain field value interpretation.

mcjaeger:
I tried to make a start for providing the content here:

https://github.com/sw360/sw360portal/wiki/User-Fields-and-Terms

PR #601, opened 2017-09-29
feat(ui): add tooltips to sw360 enum values

maierthomas:
Show tooltip text for the sw360 enum values in datatables and tables.

The tooltips ca be configured in a CSS file
- sw360-portlet/webapp/css/tooltips.css

Closes sw360/sw360portal#394

### DIFF
--- a/frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/tags/DisplayEnumInfo.java
+++ b/frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/tags/DisplayEnumInfo.java
@@ -1,0 +1,71 @@
+/*
+ * Copyright Siemens AG, 2017. Part of the SW360 Portal Project.
+ *
+ * SPDX-License-Identifier: EPL-1.0
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.eclipse.sw360.portal.tags;
+
+import org.apache.thrift.TEnum;
+import org.eclipse.sw360.datahandler.common.ThriftEnumUtils;
+import org.eclipse.sw360.portal.common.PortalConstants;
+
+import javax.servlet.jsp.JspException;
+import javax.servlet.jsp.JspWriter;
+import javax.servlet.jsp.tagext.SimpleTagSupport;
+import java.io.IOException;
+import java.util.Iterator;
+
+import static com.google.common.base.Predicates.instanceOf;
+import static com.google.common.collect.Iterables.all;
+
+/**
+ * @author thomas.maier@evosoft.com
+ */
+public class DisplayEnumInfo extends SimpleTagSupport {
+
+    private static final String SW360_INFO_IMG_PATH = "/sw360-portlet/images/ic_info.png";
+    private static final String SW360_INFO_IMG = "infopic";
+
+    private Class type;
+
+    private Iterable<? extends TEnum> options;
+
+    public void setType(Class type) {
+        this.type = type;
+    }
+
+    @SuppressWarnings("unchecked")
+    public void setOptions(Iterable options) throws JspException {
+        if (!all(options, instanceOf(TEnum.class))) {
+            throw new JspException("given type options are not of class TEnum");
+        }
+
+        this.options = (Iterable<? extends TEnum>) options;
+    }
+
+    public void doTag() throws JspException, IOException {
+        if (options != null) {
+            printEnumValuesInfo(options.getClass());
+        } else if (type != null) {
+            printEnumValuesInfo(type);
+        } else {
+            throw new JspException("you must select either a TEnum type or a collection of values");
+        }
+    }
+
+    private void printEnumValuesInfo(Class enumClass) throws IOException {
+        JspWriter jspWriter = getJspContext().getOut();
+
+        String result = "<span class='" + PortalConstants.TOOLTIP_CLASS__CSS + " "
+                + PortalConstants.TOOLTIP_CLASS__CSS + "-" + enumClass.getSimpleName() + "'>"
+                + "<img class='" + SW360_INFO_IMG + "' src='" + SW360_INFO_IMG_PATH + "'></span>";
+
+        jspWriter.print(result);
+
+    }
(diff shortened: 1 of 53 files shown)


mcjaeger:
@adrian-evo if you would have time, mind a test?

adrian-evo:
Noticed two issues so far:

1. Clearing state field is broken on edit release, it contains <span class=...
2. Moderation tab is broken

clearing

mod tab

maierthomas:
1. changed to: bare output for the clearing state in the release edit mode.
2. fixed moderation tab

mcjaeger:
@adrian-evo if you would have time, mind a re - test of the non working parts?

adrian-evo:
The issues are fixed, however I don't see what is the change, i.e. there is no (i) icon that displays a tooltip. Can you show a sample screenshot about how it is supposed to work?

frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/tags/DisplayEnumInfo.java
```diff
@@ -0,0 +1,71 @@
...
+
+import static com.google.common.base.Predicates.instanceOf;
+import static com.google.common.collect.Iterables.all;
+
+/**
+ * @author thomas.maier@evosoft.com
+ */
+public class DisplayEnumInfo extends SimpleTagSupport {
+
+    private static final String SW360_TOOLTIP = "sw360-tt";
```
alexbrdn: same constant as in DisplayEnum.class. Also, when you pull it out, it's probably better to rename it to more descriptive `SW360_TOOLTIP_CSS_CLASS` or something like that

frontend/sw360-portlet/src/main/webapp/html/components/includes/components/vulnerabilities.jspf
```diff
@@ -91,7 +91,7 @@
                 </core_rt:when>
                 <core_rt:otherwise>
                         '<div title="<sw360:out value='${vulnerabilityVerificationTooltips.get(vulnerability.externalId).get(vulnerability.intReleaseId)}'/>">'+
-                        '<sw360:DisplayEnum value="${vulnerabilityVerifications.get(vulnerability.externalId).get(vulnerability.intReleaseId)}"/>'+
+                        "<sw360:DisplayEnum value="${vulnerabilityVerifications.get(vulnerability.externalId).get(vulnerability.intReleaseId)}"/>"+
```
alexbrdn: why did you change single quotes to double quotes in so many places? I don't see the reason.

frontend/sw360-portlet/src/main/webapp/html/projects/includes/projects/basicInfo.jspf
```diff
@@ -33,6 +33,8 @@
                    value="<sw360:out value="${project.name}"/>"/>
             <img class="infopic" src="<%=request.getContextPath()%>/images/ic_info.png"
                  title="Name of the project"/>
+
```
alexbrdn: unnecessary line breaks

frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/tags/DisplayEnumInfo.java
maierthomas: Moved to Portal Constants and renamed it to TOOLTIP_CLASS__CSS

frontend/sw360-portlet/src/main/webapp/html/projects/includes/projects/basicInfo.jspf
maierthomas: Removed unnecessary line breaks

frontend/sw360-portlet/src/main/webapp/css/tooltips.css
```diff
@@ -0,0 +1,723 @@
...
+.sw360-tt-AttachmentType-LICENSE_AGREEMENT:hover:after {
+    content: "A ruling license agreement for this item, note that this could be for commercial software for example";
+}
+
+.sw360-tt-AttachmentType-SCREENSHOT:hover:after {
+    content: "Screenshot, usually screenshot of the Website with licensing information";
+}
+
+.sw360-tt-AttachmentType-README_OSS:hover:after {
+    content: "tbd";
```
maxhbr: tbd?

maierthomas: removed text tbd and hide this tooltip

frontend/sw360-portlet/src/main/webapp/html/components/includes/components/vulnerabilities.jspf
maierthomas: To avoid problems if single quotes are used in the DisplayEnum function.
Earlier I generated a span element with single quotes in the class attribute.
As a result problems occurred and I changed the single quotes to double quotes.

Currently the DisplayEnum function uses double quotes for the class value.

alexbrdn: Why didn't you use single quotes for the class attribute in DisplayEnum? I mean, it's not that important, but that would contain the change to DisplayEnum, instead of spreading it to multiple files, would it not?

maxhbr: The correct solution would be to escape the result of `DisplayEnum`?

alexbrdn: Right

maierthomas: In my opinion it doesn't make to escape the DisplayEnum result here.
The enum values itself can not have single or double quotes in their names.
Also the tooltip text is defined in the CSS file and not in javascript.

In case of double quotes in the tooltip text you need to escape them in the CSS file.

maxhbr: until someone toggles the qouting in `DisplayEnum` or maybe use both styles of qouting. Escaping is the safe way.

mcjaeger:
@alexbrdn @maxhbr we discussed to create an issue for the escaping because it seems like for other tags the issue

Issue #559, opened 2017-09-06
License generation shall be clean


mcjaeger:
The license texts contain some times some characters which are confusing in a license documentation output, most prominently, there are HTML escaped characters starting with the ampers and character.

Formatting issues that should be cleaned up before generation:

* eliminate all characters from the source files (SPDX, CLI) which shall not belong in the text
* if the output is text, the HTML escaped characters shall be just written as their text value, for HTML output, according rendering.
* if there are HTML style tags (e.g. for bold typing), they should be ignored in the text output, but rendered in the HTML output.
* if there are HTML formatting tags, they should be rendered for both cases (new line, paragraph, etc)

alexbrdn:
Note that bullet point 2 
> * if the output is text, the HTML escaped characters shall be just written as their text value, for HTML output, according rendering.

is solved by the combination of #586 and #656

PR #656, opened 2017-10-13
feat(licenseinfo): clean license texts in license info generation

ChristophNiehoff:
tiny modification to the velocity template for plain text generation.
Removed xml-escaping.

### DIFF
--- a/backend/src/src-licenseinfo/src/main/resources/textLicenseInfoFile.vm
+++ b/backend/src/src-licenseinfo/src/main/resources/textLicenseInfoFile.vm
@@ -90,7 +90,7 @@ Open Source Software Licenses
 #foreach($license in $allLicenseNamesWithTexts)
 #set($licenseName = "<no license name available>")
 #if($license.licenseName)
-#set($licenseName = $esc.xml($license.licenseName))
+#set($licenseName = $license.licenseName)
 #end## if($license.licenseName)
 ##
 #set($licenseText = "<no license text available>")


ChristophNiehoff:
@mcjaeger At least for my example files this results in nice looking license infos. Could you check if this small fix is sufficient?

adrian-evo:
Tried with the attached file and I still can see some instances of &gt; in the generated txt file.

[CLI_Apache_Thrift-0.9.3.xml.zip](https://github.com/sw360/sw360portal/files/1394770/CLI_Apache_Thrift-0.9.3.xml.zip)

alexbrdn:
I believe this fix is insufficient. It covers only the second bullet point from #559.

ChristophNiehoff:
@alexbrdn In a private discussion we decided, that we concentrate on unescaping the HTML-escaped characters for the time being. And right now, we do not include rendering the HTML tags.

alexbrdn:
Well, if it was private, how am I supposed to know?

Anyways... in light of what I just learned, I think the "Closes" tag in the PR description is incorrect. We don't want to auto-close the issue when this PR is merged, as it fixes only a part of the issue.

Apart from that, I can confirm that it works.

@adrian-evo, I think you made a mistaken assumption as to what is required of this fix. This PR shouldn't remove all instances of angle brackets from text (there are some in the original). What it should do is to generate text file without escaping the angle brackets and such as if it was generating an xml file.

ChristophNiehoff:
@alexbrdn  Ok, yes. I agree. Removed the "Closes" tag.

maxhbr:
The closes tag is still in the commit message:

ChristophNiehoff:
@maxhbr You are faster in commenting than me in committing... ;)

Anyways, during rebase I realized that most part of this fix was already included in the already merged PR  https://github.com/sw360/sw360portal/pull/586 (authored by me). This makes the current PR kind of dispensable. However, the question is still open, whether this is sufficient.

alexbrdn:
I believe it is sufficient to remove the xml escaping when generating text document - just as you have done.

PR #660, opened 2017-10-17
feat(lics): also import and export external IDs

maxhbr:
also add (optional) external IDs for Licenses and TODOs (only used while importing and exporting)

### DIFF
--- a/libraries/commonIO/pom.xml
+++ b/libraries/commonIO/pom.xml
@@ -36,5 +36,10 @@
             <groupId>org.hamcrest</groupId>
             <artifactId>hamcrest-all</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.google.code.gson</groupId>
+            <artifactId>gson</artifactId>
+            <version>2.8.0</version>
+        </dependency>
     </dependencies>
 </project>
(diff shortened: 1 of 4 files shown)


mcjaeger:
how to test it?

* create some licenses (maybe A B C) and then some todos (maybe 1 2 3)
* then export the licesenses (in the admin section)
* Then open the *.lics archive (using unzipper tool)
* edit the text file with the todos
* change some external Ids there for todo (maybe go to TODO 1 and then add some external id)
* then compress again the folder with the files inside
* rename extension to *.lics again
* then clean up the database (remove all licenses and todos)
* try to upload / import to an empty database.

-> then look into the couchdb if the license todo has a externalid

adrian-evo:
Tested and seems to work partially for License IDs, however importing TODOs is not working due to #625 

Even if a single license and a single TODO is exported and imported back, the TODO is not imported.

maxhbr:
@adrian-evo You need to add the todos via import and not via the ui. Then it will work.

One might be able to fix that by adding the sw360 internal id as external id with key `"sw360"` to the export.

adrian-evo:
Imported the following file as lics, containing a single license and a single todo, both with external ids. The external ids are displayed fine in database, including for todo, but the todo cannot be seen on the UI.

[LicensesBackup4.zip](https://github.com/sw360/sw360portal/files/1449886/LicensesBackup4.zip)

PR #662, opened 2017-10-18
feat(ui): Make Liferay session expire after 30 minutes

bodetc:
All is in the title

### DIFF
--- a/frontend/sw360-portlet/src/main/resources/portal.properties
+++ b/frontend/sw360-portlet/src/main/resources/portal.properties
@@ -11,3 +11,12 @@
 #auto.login.hooks=org.eclipse.sw360.portal.users.SSOAutoLogin
 login.events.post=org.eclipse.sw360.portal.users.LoginAction
 servlet.service.events.pre=org.eclipse.sw360.portal.hooks.BuildInfoForVelocityProviderHook
+
+
+# Specify the number of minutes before a session expires. This value is
+# always overridden by the value set in web.xml.
+session.timeout=30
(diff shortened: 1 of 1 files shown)


frontend/sw360-portlet/src/main/resources/portal.properties
```diff
@@ -10,4 +10,13 @@
...
 #auto.login.hooks=org.eclipse.sw360.portal.users.SSOAutoLogin
 login.events.post=org.eclipse.sw360.portal.users.LoginAction
-servlet.service.events.pre=org.eclipse.sw360.portal.hooks.BuildInfoForVelocityProviderHook
\ No newline at end of file
+servlet.service.events.pre=org.eclipse.sw360.portal.hooks.BuildInfoForVelocityProviderHook
+
+
+# Specify the number of minutes before a session expires. This value is
+# always overridden by the value set in web.xml.
+session.timeout=30
```
maierthomas: Why do we need to set the session timeout in this specific file (portal.properties)?

Currently in our liferay configuration the session timeout is already configured.
This timeout is configured in the web.xml (ROOT/WEB-INF) 
```
	<session-config>
		<session-timeout>30</session-timeout>
	</session-config>
```

Which configuration file is now the "master"?
- web.xml or portal.properties?

bodetc:
Here is the information regarding precedence for the timeout: https://web.liferay.com/community/wiki/-/wiki/Main/Session+Timeout
> The web.xml setting takes precedence over portal.properties

Apparently, we do not have this setting on our productive instance. You can test it by removing the entries from your local web.xml

PR #663, opened 2017-10-18
feat(vulnerabilities): Make CVE search configurable

bodetc:
Allows for configuring the CVE search using sw360.properties

### DIFF
--- a/backend/src/src-cvesearch/src/main/java/org/eclipse/sw360/cvesearch/datasource/heuristics/SearchLevels.java
+++ b/backend/src/src-cvesearch/src/main/java/org/eclipse/sw360/cvesearch/datasource/heuristics/SearchLevels.java
@@ -12,36 +12,45 @@
  */
 package org.eclipse.sw360.cvesearch.datasource.heuristics;
 
+import org.apache.log4j.Logger;
 import org.eclipse.sw360.cvesearch.datasource.CveSearchApi;
 import org.eclipse.sw360.cvesearch.datasource.CveSearchGuesser;
 import org.eclipse.sw360.cvesearch.datasource.matcher.Match;
+import org.eclipse.sw360.datahandler.common.CommonUtils;
 import org.eclipse.sw360.datahandler.thrift.components.Release;
-import org.apache.log4j.Logger;
 
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
+import java.util.Properties;
 import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import static java.util.Collections.singletonList;
 import static org.eclipse.sw360.datahandler.common.CommonUtils.nullToEmptyString;
-import static java.util.Collections.*;
 
 public class SearchLevels {
 
-    private static final String CPE_PREFIX        = "cpe:2.3:";
-    private static final String OLD_CPE_PREFIX    = "cpe:/";
-    private static final String CPE_WILDCARD      = ".*";
+    private static final Logger log = Logger.getLogger(SearchLevels.class);
+
+    private static final String CPE_PREFIX = "cpe:2.3:";
+    private static final String OLD_CPE_PREFIX = "cpe:/";
+    private static final String CPE_WILDCARD = ".*";
     private static final String CPE_NEEDLE_PREFIX = CPE_PREFIX + ".:";
-    public static final int DEFAULT_VENDOR_THRESHOLD  = 1;
-    public static final int DEFAULT_PRODUCT_THRESHOLD = 0;
-    public static final int DEFAULT_CUTOFF            = 6;
 
-    private Logger log = Logger.getLogger(SearchLevels.class);
+    private static final String PROPERTIES_FILE_PATH = "/sw360.properties";
+    private static final String VENDOR_THRESHOLD_PROPERTY = "cvesearch.vendor.threshold";
+    private static final String PRODUCT_THRESHOLD_PROPERTY = "cvesearch.product.threshold";
+    private static final String CUTOFF_PROPERTY = "cvesearch.cutoff";
 
-    private List<SearchLevel> searchLevels;
+    private static final int DEFAULT_VENDOR_THRESHOLD = 1;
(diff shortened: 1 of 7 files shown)


mcjaeger:
maybe, if you find it use maybe some info to the threshold value could be added. (because for the threshold it is not obvious what the threshold does) but this is not review / relevant IMHO.

backend/src/src-cvesearch/src/main/java/org/eclipse/sw360/cvesearch/datasource/heuristics/SearchLevels.java
```diff
@@ -77,48 +93,49 @@ private void setup(CveSearchApi cveSearchApi, int vendorThreshold, int productTh
...
+
     private void addCPESearchLevel() {
         Predicate<Release> isPossible = r -> r.isSetCpeid() && isCpe(r.getCpeid().toLowerCase());
         searchLevels.add(r -> {
-            if(isPossible.test(r)){
+            if (isPossible.test(r)) {
                 return singletonList(new NeedleWithMeta(cleanupCPE(r.getCpeid()), "CPE"));
             }
-            return EMPTY_LIST;
+            return ImmutableList.of();
```
alexbrdn: Why do you think this is better than `Collections.EMPTY_LIST`? It doesn't look like a necessary change.

backend/src/src-cvesearch/src/main/java/org/eclipse/sw360/cvesearch/datasource/heuristics/SearchLevels.java
```diff
@@ -77,48 +93,49 @@ private void setup(CveSearchApi cveSearchApi, int vendorThreshold, int productTh
...
+                            throw new UncheckedIOException(e);
+                        }
+                    });
         } catch (UncheckedIOException ue) {
             throw ue.getIOExceptionCause();
         }
     }
 
-    class UncheckedIOException extends RuntimeException{
+    class UncheckedIOException extends RuntimeException {
```
alexbrdn: Here and at multiple other places in this PR: unnecessary change in formatting. We have agreed not to do that. See https://github.com/sw360/sw360portal/wiki/Dev-DoD-and-Style

Please correct this first as it makes very hard to see what has been actually changed in the functionality

backend/src/src-cvesearch/src/test/resources/sw360.properties
```diff
@@ -0,0 +1,11 @@
...
+# Copyright (c) Bosch Software Innovations GmbH 2016.
+#
+# All rights reserved. This configuration file is provided to you under the
+# terms and conditions of the Eclipse Distribution License v1.0 which
+# accompanies this distribution, and is available at
+# http://www.eclipse.org/org/documents/edl-v10.php
+
+cvesearch.default.vendor.threshold=1
+cvesearch.default.product.threshold=0
+cvesearch.default.cutoff=6
```
alexbrdn: missing newline at EOF

backend/src/src-cvesearch/src/main/java/org/eclipse/sw360/cvesearch/datasource/heuristics/SearchLevels.java
bodetc: It avoids an unchecked cast. It is definitely not necessary, but do we have a guideline to make only necessary changes? I am used to work with a Sonar instance running on the code, so I tend to correct the "warnings" when I see them.

backend/src/src-cvesearch/src/test/resources/sw360.properties
bodetc: Doesn't it count as formatting? I fixed this anyway :)

backend/src/src-cvesearch/src/main/java/org/eclipse/sw360/cvesearch/datasource/heuristics/SearchLevels.java
bodetc: I have corrected the feature in my IDE, only changed lines will be formatted. I don't see the point to revert the line that were changed by mistake back to an inconsistant formatting.

backend/src/src-cvesearch/src/main/java/org/eclipse/sw360/cvesearch/datasource/heuristics/SearchLevels.java
alexbrdn: No, no, I do not suggest to only make the absolutely necessary changes. I only wondered, what moved you to reduce the readability of the code. IMO, `EMPTY_LIST` is much easier to parse at a glance.
What about using `Collections.emptyList()`? Looks like a win-win to me.

backend/src/src-cvesearch/src/test/resources/sw360.properties
alexbrdn: Yes, it is formatting, obviously. But that does not count as changing someone else's formatting.

backend/src/src-cvesearch/src/main/java/org/eclipse/sw360/cvesearch/datasource/heuristics/SearchLevels.java
alexbrdn: Thing is, it's not "the line". It's a lot of lines. I'm having a hard time reviewing what you actually changed. I can't and won't make you do it, but it would be oh so kind of you if you did.

bodetc: I reverted most of them, it should be more readable now :)

backend/src/src-cvesearch/src/main/java/org/eclipse/sw360/cvesearch/datasource/heuristics/SearchLevels.java
bodetc: I changed both occurences to Collections.emptyList()

backend/src/src-cvesearch/src/test/resources/sw360.properties
bodetc: done.

Issue #669, opened 2017-10-24
License info header contains html tags on moderation page


adrian-evo:
Created a new project, then a moderation request for any field, e.g. homepage. When opening the moderation request, the license info header contains <span title, etc. tags. If default text was modified previously, the header contains <br.

This is only on moderation page, the license text is not affected.

![moderation](https://user-images.githubusercontent.com/19324942/31927584-75ec057c-b89d-11e7-83ca-6568010b7f1e.PNG)

![mod2](https://user-images.githubusercontent.com/19324942/31927600-81a7d562-b89d-11e7-87ba-49942d6a9318.PNG)

PR #670, opened 2017-10-24
fix(moderation): No HTML tags in moderation requests for license info…

ChristophNiehoff:
… headers of projects

Closes sw360/sw360portal#669

### DIFF
--- a/frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/tags/DisplayProjectChanges.java
+++ b/frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/tags/DisplayProjectChanges.java
@@ -300,17 +300,19 @@ public class DisplayProjectChanges extends UserAwareTag {
     }
 
     private Project prepareLicenseInfoHeaderTextInProject(Project project) {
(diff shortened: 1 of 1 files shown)


ChristophNiehoff:
This is a fix to issue https://github.com/sw360/sw360portal/issues/669 .

The problem was that in the `<sw360:DisplayProjectChanges>` tag (for displaying the changes of the project in a moderation request) the original project objects were erroneously overwritten.

adrian-evo:
Works as expected.

frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/tags/DisplayProjectChanges.java
```diff
@@ -300,17 +300,19 @@ private boolean ensureSomethingTodoAndNoNullReleaseIdUsage() {
     }
 
     private Project prepareLicenseInfoHeaderTextInProject(Project project) {
+        Project modifiedProject = project.deepCopy();
```
maierthomas: Instead of modifiedProject we could use newProject as in other deep copy variable names.

Issue #327, opened 2017-01-18
HTML-Escaping Descriptions Of Components And Projects


alexbrdn:
Descriptions of components and projects are not escaped correctly on homepage portlets. Specifically, two following problems were discovered on the Home page:
- MyComponents table can't handle an apostrophe in a component description.
- MyProjects table can't handle newline chars in a project description.

PR #671, opened 2017-10-26
fix(homepage): Escape description value for title attributes on homepage

maierthomas:
This PR resolves the syntax error for description fields on the homepage.
- Escape HTML title attribute correctly (If description of components or projects more than 30 characters)

closes sw360/sw360portal#327


frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/tags/OutTag.java
```diff
@@ -86,6 +88,11 @@ protected String escapeInSingleQuote(String value) {
         return value.replaceAll("'", quoteReplacement("\\\'"));
     }
 
+    private String prepareTitleAttribute(String value) {
```
ChristophNiehoff: I didn't test this thoroughly, but since the problem stems from the JS code, I think a simple
`StringEscapeUtils.escapeJavaScript(value)` might be enough.  Then you could get rid of  `escapeInDoubleQuote` and `prepareTitleAttribute` alltogether.  I quickly implemented it and it seemed to work.
This might be more foolproof than just replacing a few characters.

However, I am not sure, if this leads to problems elsewhere....

ChristophNiehoff: It seems that this approach leads to problems with the `<sw360:ProjectName>`  as this inherits from the `<sw360:out>` tag....

But there also is an `<sw360:DisplayDescription>` tag.  Maybe one should use this on the homepage.

ChristophNiehoff:
On the component view page, the components names are not escaped. Maybe we can also change this in this PR?    I did not check other pages though...

frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/tags/OutTag.java
```diff
@@ -86,6 +88,11 @@ protected String escapeInSingleQuote(String value) {
...
+        // Escape attribute and replace line breaks in span title element otherwise a syntax error occurs
+        return escapeAttributeValue(value).replaceAll("[\r\n]+", "&#013;&#010;");
```
alexbrdn: speaking of escaping... I'm not sure this replaceAll is adequate. Looks like a string "\n\r\n" will be converted to "&#013;&#010;&#013;&#010;&#013;&#010;"

frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/tags/OutTag.java
maierthomas: What about to remove the ´<sw360:DisplayDescription>´ tag? This tag is not really necessary and duplicated code. Instead of ´<sw360:DisplayDescription>´ we can use the normal ´<sw360:out>´ tag. 
I also removed the ´ escapeInDoubleQuote´ and ´ escapeInSingleQuote´ functions. I think also its better to use ´SringEscapeUtils.escapeJavaScript()´

The issue regarding ProjectName and ComponentName is solved by decoding HTML entities.

Id did a lot of tests and everything looks quite good now.

frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/tags/OutTag.java
maierthomas: Added missing ´sw360:out´ for:
- component name (view)
- release name (clearing status)
- linked releases (vendor, name, version)

maierthomas:
Add missing `sw360:out´ tags:
- name (component view)
- vendor, name, version (linked releases)
- release name (clearing status)

frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/tags/OutTag.java
```diff
@@ -73,38 +72,35 @@ public int doStartTag() throws JspException {
...
     }
 
-    //TODO remove unneeded...
-    protected String escapeInSingleQuote(String value) {
-        return value.replaceAll("'", quoteReplacement("\\\'"));
+    public void setValue(Object value) {
+        this.value = value;
     }
 
     public void setDefault(String def) {
```
ChristophNiehoff: The `default` string is not used anymore? Can it be removed?

frontend/sw360-portlet/src/main/webapp/WEB-INF/customTags.tld
```diff
@@ -738,6 +738,17 @@
         </attribute>
     </tag>
     <tag>
+        <name>ComponentName</name>
```
ChristophNiehoff: Looking at `customTags.tld`, it seems that almost all tags start with the prefix "Display". Maybe we should promote this to a strict convention for this project and call this tag `DisplayComponetName` (and accordingly for the other tags)?
Maybe the `out`-tag can be an exception as it is a general purpose thing....  ?

frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/tags/OutTag.java
ChristophNiehoff: Sounds all good to me! :)

adrian-evo:
Works as expected.

frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/tags/OutTag.java
maierthomas: The default string (def) is used in the super method call, so it should not be removed.

frontend/sw360-portlet/src/main/webapp/WEB-INF/customTags.tld
maierthomas: Well in our customTags we have some extra tags to show only the Name of the object.
e.g. tags: RelaseName, ProjectName, LicenseName.
As a result I added this for the component.

PR #676, opened 2017-11-07
feat(ui): Define new LAR files to remove BDP Import from default configuration

bodetc:
There is now to privates pages LAR files. The Import portlet is now in a separate LAR files. The files can be imported one after the other.


mcjaeger:
I am not sure and would like to ask for feedback - how about putting the bdp-specific lar file into the bdp project (and create a folder called `frontend`) there?

bodetc:
@mcjaeger That was our initial assumption. But the frontend code is not modular and the BDP Portlet itself is in the main repository, all in the "frontend" project. This is why we decided to place the LAR file there.

mcjaeger:
just testing is needed. I think the "problem" can be seen here: 

https://github.com/sw360/sw360portal/search?utf8=%E2%9C%93&q=bdp&type=

shall create an issue and merge this PR?

mcjaeger:
@maxhbr we remember faintly that you said in yesterdays telco, you have tested this?

Issue #349, opened 2017-02-23
 ECC fields - show ECC status also on project level 


mcjaeger:
The ECC status is there for components, however, for project or product responsibles, it would be helpful to have the ECC overview for all releases of a product or project. It should be similar to the clearing status view, to have also the `ECC Status` view.

Listed columns should be same as on the general ECC listing page.

Also this should show the same export spreadsheet button at the bottom right corner of the list  (adding just the button to trigger the same export as for clearing status).

PR #677, opened 2017-11-07
feat(project): add ECC status information to projects

maierthomas:
- add new tab ECC Status to projects (same fields as in ECC listing page)
- export release spreadsheet will return the same excel sheet as from Clearing Status tab

closes sw360/sw360portal#349

### DIFF
--- a/frontend/sw360-portlet/src/main/webapp/html/projects/includes/projects/eccStatus.jspf
+++ b/frontend/sw360-portlet/src/main/webapp/html/projects/includes/projects/eccStatus.jspf
@@ -1,0 +1,100 @@
+<%--
+  ~ Copyright Siemens AG, 2017. Part of the SW360 Portal Project.
+  ~
+  ~ SPDX-License-Identifier: EPL-1.0
+  ~
+  ~ All rights reserved. This program and the accompanying materials
+  ~ are made available under the terms of the Eclipse Public License v1.0
+  ~ which accompanies this distribution, and is available at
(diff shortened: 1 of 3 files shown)


frontend/sw360-portlet/src/main/webapp/html/projects/includes/projects/eccStatus.jspf
```diff
@@ -0,0 +1,100 @@
...
+                <th width="20%"></th>
+                <th width="20%"></th>
+                <th width="10%"></th>
+            </tr>
+            </tfoot>
+        </table>
+        <core_rt:if test="${inProjectDetailsContext}">
+            <div class="clear-float"></div>
+            <span class="pull-right">
+                <input type="button" class="addButton" id="exportReleasesSpreadsheetECCButton" value="Export Spreadsheet"/>
```
bodetc: I get an Internal Server Error when clicking this button, cause by a "NullPointerException" while generating the XLSX. It seems to be due to a missing font. Does anything needs to be configured?

maierthomas: I cannot reproduce this behavior.
Alex and me could execute the export spreadsheet function successfully.
@mcjaeger please verify the export spreadsheet function on both pagaes (clearing status / ecc status)

mcjaeger: no, this null pointer excpetion I did not see. Generation works fine for me.

PR #661, opened 2017-10-18
feat(bdpimport): Allow user to search for BDP projects instead of loading them all

bodetc:
Must be considered with PR https://github.com/sw360/sw360bdpImportService/pull/4 from bdpImportService

PR #680, opened 2017-11-15
feat(bdpImport): Use of proper error message when importing projects

bodetc:
Instead of a list of IDs, now a error message is shown for each failed ID with a reason for the failure.

Backend PR:  https://github.com/sw360/sw360bdpImportService/pull/5
Depends on PR: https://github.com/sw360/sw360portal/pull/661

(note that before the message was like 'there was some problem', now it shall display which projects etc)

### DIFF
--- a/frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/common/PortletUtils.java
+++ b/frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/common/PortletUtils.java
@@ -307,9 +307,9 @@ public class PortletUtils {
         JSONArray jsonNewIds = JSONFactoryUtil.createJSONArray();
         JSONArray jsonUpdatedIds = JSONFactoryUtil.createJSONArray();
 
-        updateStatus.getStatusToVulnerabilityIds().get(UpdateType.FAILED).forEach(id -> jsonFailedIds.put(id));
-        updateStatus.getStatusToVulnerabilityIds().get(UpdateType.NEW).forEach(id -> jsonNewIds.put(id));
-        updateStatus.getStatusToVulnerabilityIds().get(UpdateType.UPDATED).forEach(id -> jsonUpdatedIds.put(id));
+        updateStatus.getStatusToVulnerabilityIds().get(UpdateType.FAILED).forEach(jsonFailedIds::put);
+        updateStatus.getStatusToVulnerabilityIds().get(UpdateType.NEW).forEach(jsonNewIds::put);
+        updateStatus.getStatusToVulnerabilityIds().get(UpdateType.UPDATED).forEach(jsonUpdatedIds::put);
 
         responseData.put(PortalConstants.UPDATE_VULNERABILITIES__FAILED_IDS, jsonFailedIds);
         responseData.put(PortalConstants.UPDATE_VULNERABILITIES__NEW_IDS, jsonNewIds);
--- a/frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/portlets/projectimport/ProjectImportPortlet.java
+++ b/frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/portlets/projectimport/ProjectImportPortlet.java
@@ -224,8 +224,9 @@ public class ProjectImportPortlet extends Sw360Portlet {
 
     private void importBdpProjects(User user, List<String> selectedIds, JSONObject responseData, RemoteCredentials remoteCredentials) throws PortletException, IOException {
         ImportStatus importStatus = importDatasources(selectedIds, user, remoteCredentials);
-        JSONArray jsonFailedIds = JSONFactoryUtil.createJSONArray();
+        JSONObject jsonFailedIds = JSONFactoryUtil.createJSONObject();
         JSONArray jsonSuccessfulIds = JSONFactoryUtil.createJSONArray();
+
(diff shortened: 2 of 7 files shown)


frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/common/PortalConstants.java
```diff
@@ -320,49 +320,49 @@
...
+
+    // Project Import: credentials/search string
+    public static final String PROJECT_IMPORT_USERNAME = "projectImport_import_user";
+    public static final String PROJECT_IMPORT_PASSWORD = "projectImport_import_pass";
+    public static final String PROJECT_IMPORT_SERVER_URL = "projectImport_import_url";
+    public static final String PROJECT_IMPORT_PROJECT_NAME = "projectImport_import_project_name";
+
+    // Project Import: user actions
+    public static final String PROJECT_IMPORT_USER_ACTION__IMPORT = "projectImport_import_action";
+    public static final String PROJECT_IMPORT_USER_ACTION__IMPORT_BDP = "projectImport_import_bdp_data";
```
mcjaeger: why not calling this `PROJECT_IMPORT_USER_ACTION__IMPORT_DATA`

frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/common/PortalConstants.java
```diff
@@ -320,49 +320,49 @@
...
+    public static final String PROJECT_IMPORT_USER_ACTION__DISCONNECT = "projectImport_disconnect";
+    public static final String PROJECT_IMPORT_USER_ACTION__NEWIMPORTSOURCE = "projectImport_select_new_import_source";
+    public static final String PROJECT_IMPORT_USER_ACTION__UPDATEIMPORTABLES = "projectImport_update_importables";
+
+    // Project Import: response fields
+    public static final String PROJECT_IMPORT_RESPONSE__SUCCESS = "projectImport_response_bdp_import_success";
+    public static final String PROJECT_IMPORT_RESPONSE__FAILURE = "projectImport_response_bdp_import_failure";
+    public static final String PROJECT_IMPORT_RESPONSE__GENERAL_FAILURE = "projectImport_response_bdp_import_general_failure";
+
```
mcjaeger: maybe removing `bdp` could be also OK

frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/portlets/projects/ProjectImportPortlet.java
```diff
@@ -128,12 +131,12 @@ private boolean isImportSuccessful(ImportStatus importStatus) {
         return (importStatus.isSetRequestStatus() && importStatus.getRequestStatus().equals(RequestStatus.SUCCESS) && importStatus.getFailedIds().isEmpty());
     }
 
-    private ImportStatus importDatasources(List<String> toImport, User user, RemoteCredentials remoteCredentials)  {
+    private ImportStatus importDatasources(List<String> toImport, User user, RemoteCredentials remoteCredentials) {
         ImportStatus importStatus = new ImportStatus();
         try {
             importStatus = bdpImportClient.importDatasources(toImport, user, remoteCredentials);
             if (!isImportSuccessful(importStatus)) {
```
mcjaeger: if the thrift file could be renamed this could be also renamed at the same occasion.

mcjaeger:
Since you are adjusting labels / names, such as

` SESSION_IMPORT_USER` to `PROJECT_IMPORT_USERNAME`

How about eliminating also the references to BDP from the sw360portal code base? for example. it does not make sense to have the file "bdpImport.thrift" (https://github.com/sw360/sw360portal/blob/master/libraries/lib-datahandler/src/main/thrift/bdpimport.thrift), but it could be renamed to "projectImport.thrift"

At the same time, a variables named with `bdp` could be changed too ... do not get me wrong on this, I know that this PR is not targeting cleaning up the code. I am just proposing this because you are changing these lines anyway and then the changed lines could also remove the `bdp` references, because it would be cleaner if the project import functionality is actually called project import regardless of the source.

mcjaeger:
hi, I am not sure about the error message improvement. Got the following from screenshot (see below), what I did it is:
* deployed bdpimporter on dev/654-bdp-error-messages
* deployed sw360portal branch bsinno:dev/654-bdp-error-messages

I keep on testing because the log says that the importing of the license crashes and I guess this does not work with the plain user:
```
 2017-11-29 15:06:48 ERROR ThriftExchange:249 - Could not add License for user with email=[user1@sw360.org]:org.apache.thrift.TApplicationException: addLicenses failed: unknown result
```

screen shot 2017-11-29 at 16 19 35

mcjaeger:
Update: with an admin user it seems to work. I guess the message is also improved. So, what do we do about the "License cannot be created" problem ... I think the import of projects is maybe not for every sw360 user but only for those who can create licenses anyways? As such the import menu item could be made visible only to those during deployment (requires updated lar though)
screen shot 2017-11-29 at 16 28 42

Attached a screenshot of the out that you maybe anticipated.

bodetc:
The "License cannot be created" is unfortunately hard to catch, as it comes as a exception in the `addProject` method of `projects.thrift`. It would require us to also refactor that method, and do one more breaking change in the thrift API.

PR #713, opened 2018-01-12
chore(rest): Update for REST API documentation

maierthomas:
- changed subtitle and hierarchy level
- added status code
- added chapter 'Token'
- added some fields for components and projects

### DIFF
--- a/rest/resource-server/src/docs/asciidoc/api-guide.adoc
+++ b/rest/resource-server/src/docs/asciidoc/api-guide.adoc
@@ -8,12 +8,12 @@
 //
 
 = sw360 REST API Guide
-BETA; Version {project-version}
+Documentation; Version {project-version}
 :doctype: book
 :icons: font
 :source-highlighter: highlightjs
 :toc: left
-:toclevels: 4
+:toclevels: 3
 :sectlinks:
 :numbered:
 
@@ -42,11 +42,11 @@ use of HTTP verbs.
 | `PUT`
 | Used to update an existing resource (complete update)
 
-| `PATCH`
-| Used to update an existing resource (partial updates)
+// | `PATCH`
+// | Used to update an existing resource (partial updates)
 
-| `DELETE`
-| Used to delete an existing resource
+// | `DELETE`
+// | Used to delete an existing resource
 |===
 
 [[overview-http-status-codes]]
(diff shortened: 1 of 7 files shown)


maxhbr:
Maybe also mention that the java-web-token can be decoded on https://jwt.io/

As a result one gets:
#### Header:
```
{
  "alg": "RS256",
  "typ": "JWT"
}
```
#### Payload
```
{
  "aud": [
    "sw360-REST-API"
  ],
  "user_name": "user1@sw360.org",
  "scope": [
    "all"
  ],
  "exp": 1515779086,
  "authorities": [
    "READ",
    "WRITE"
  ],
  "jti": "699c2af1-562b-4468-990f-2612b37bf7c9",
  "client_id": "trusted-sw360-client"
}
```

## Unrelated question:
Why does the `user1` get the authority `WRITE`? I think that users no longer have write access and thus should no longer contain that information in the jwt?

maierthomas:
Yes we should think about the decode option for the JWT.
Maybe we can implement this feature to reduce the annoying and long tokens.

I just want to remember that this PR is only about the enhancement of the API documentation.
Lets clarify the second part regarding the authorities in separate call.

maxhbr:
I just want to suggest to mention https://jwt.io/ in the documention about the token. I do not want to suggest anything to implement or so

maierthomas:
Oh ok. Good information.
I was confused, I thought it needs to be implemented.

On my configuration I received for user1 the following:

```
{
  "aud": [
    "sw360-REST-API"
  ],
  "user_name": "user1@sw360.org",
  "scope": [
    "all"
  ],
  "exp": 1516023229,
  "authorities": [
    "READ"
  ],
  "jti": "073c9065-b11c-46a9-a5e4-371144c57341",
  "client_id": "trusted-sw360-client"
}
```

maxhbr:
My information is from your token in the documentation

maierthomas:
Ahhh okay...
the JWT was generated on our siemens dev system of sw360.
This system is configured as the following:
rest.write.access.usergroup=USER
Every user which belongs to usergroup USER is admin... :)
I will replace this JWT and add information about the encode decode option

maierthomas:
- added jwt.io decode information
- changed access token value
- added documentation for vulnerabilities

maierthomas:
#712 needs to be merged first!

Issue #119, opened 2016-03-15
unset project responsible/lead architect


heydenreich:
Once the lead architect or project responsible for a project is set, it is not possible to unset it. Changing to a different user is possible. Unset is possible for moderators - it should also be possible for lead architect and project responsible.

mcjaeger:
just add button to clear

mcjaeger:
(from discussion) while the project responsible might be kind of mandatory (just search the right one), this is a general issue for the single user input fields (also in components or releases) that use a radio button selection in the user search view. Once one user has been selected in this listing, there is now way to unset the radio button entirely. There should be a delete or unset option.

PR #727, opened 2018-01-25
chore(ui): add reset button in user search dialog

maierthomas:
- new button 'reset'
- removes checked attribute for input elements
- perform a new search for users

closes sw360/sw360portal#119


adrian-evo:
Works as expected.

adrian-evo:
While looking into something else, noticed that Licenses field is broken. The error on the browser is opts.$resetButton is undefined.

adrian-evo:
Works now.

Issue #758, opened 2018-02-27
REST API Documentation is missing after maven project build


maierthomas:
The REST API documentation is missing if I build the sw360 project the first time.
HTML files are not included in the resource.war package.

Steps to reproduce:
1. Clean `target` folder of project ...rest/resource/
2. run mvn package
3. Verify if the resource.war file contains the REST API documentation

PR #759, opened 2018-02-27
fix(rest): change maven plugin order to generate the documentation correctly

maierthomas:
- change maven phase for plugins `'generate-docs',` `'copy-resources'`
--> **prepare-package**

moved plugin `'pring-boot-maven-plugin'` to the end.

closes sw360/sw360portal#758


mcjaeger:
works.

Issue #451, opened 2017-06-12
Release name should be component name - and not editable


mcjaeger:
Somehow, not obvious, but release names can be changed from component name. That would make problems. It should not be editable, the release name.

alexbrdn:
I thought I'd heard from someone that that was a conscious decision (to make the name editable), but my memory is pretty dim on this.

mcjaeger:
from me? (maybe ...) but that is the summary from our usage now that in fact from the usage experience it is confusing. Maybe we need more discussion on this what a good behavior for this would be.

mcjaeger:
(from discussion) component name should be copied (and not user editable anymore) to the release name.

maierthomas:
What about existing releases with not identical names?

maierthomas:
@mcjaeger do we need to update all the existing release names with not identical names?

mcjaeger:
I think we would need a migration script for this and the migration script with dry run mode. the script should output the diverging release names and then the user can decide to go ahead.

PR #770, opened 2018-03-06
feat(release): make release names not editable

maierthomas:
- changed release edit view (component name is set to release name and is not editable)
- added migration script (_009_overwrite_release_name_with_component_name.py_)
(this scripts supports `DRY_RUN` option to show the different release/component names)

closes sw360/sw360portal#451


adrian-evo:
Works fine. Script also is working.

PR #783, opened 2018-03-15
feat: make SSOAutoLogin actually work

alexbrdn:
- make SSOAutoLogin actually work
- add possibility to map organizations from external org codes to Liferay organizations (adapted from ldapOrganizationMappingExt)
- Create Liferay organizations and users when necessary, (re-)assign user to org when necessary.

### DIFF
--- a/frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/users/OrganizationHelper.java
+++ b/frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/users/OrganizationHelper.java
@@ -1,0 +1,159 @@
+/*
+ * Copyright Bosch Software Innovations GmbH, 2016.
+ * Copyright Siemens AG, 2016-2018.
+ * Part of the SW360 Portal Project.
+ *
+ * SPDX-License-Identifier: EPL-1.0
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.eclipse.sw360.portal.users;
+
+import com.liferay.portal.NoSuchOrganizationException;
+import com.liferay.portal.kernel.bean.PortalBeanLocatorUtil;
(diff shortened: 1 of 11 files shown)


frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/users/OrganizationHelper.java
```diff
@@ -0,0 +1,161 @@
...
+
+        if (customMappingEnabled) {
+            if (matchPrefix) {
+                matcher = e -> name.startsWith(e.getKey());
+            } else { // match complete name
+                matcher = e -> name.equals(e.getKey());
+            }
+        } else {
+            // if mapping is disabled, force ignoring all mappings and default to original name
+            matcher = e -> false;
```
maxhbr: why not just: `return name`?

frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/users/UserPortletUtils.java
```diff
@@ -52,29 +68,35 @@ private static User addLiferayUser(PortletRequest request, String firstName, Str
...
 
         try {
-            ServiceContext serviceContext = ServiceContextFactory.getInstance(request);
             long[] roleIds = roleId == 0 ? new long[]{} : new long[]{roleId};
             long[] organizationIds = organizationId == 0 ? new long[]{} : new long[]{organizationId};
             long[] userGroupIds = null;
-            long currentUserId = themeDisplay.getUserId();
+            Optional<ServiceContext> serviceContextOpt = requestAdapter.getServiceContext();
+            final ServiceContext serviceContext;
+            if (!serviceContextOpt.isPresent()){
```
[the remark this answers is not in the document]
alexbrdn: I don't think it would be helpful in this situation.

frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/users/OrganizationHelper.java
alexbrdn: good point. cleaned it up

PR #784, opened 2018-03-15
chore(datahandler): remove assertion for couchdb test (lib-datahandler)

maierthomas:
- unfortunately this test fails because the blackhole ip (_192.0.2.0_) is not reachable in the network.
- this can happen in automated ci environments

```
FAILURE!
java.lang.AssertionError: 
Expected: (a string containing "timed out" or a string containing "Das Netzwerk ist nicht erreichbar")
     but: was "No route to host (Host unreachable)"
```


maxhbr:
@maierthomas that was just an internationalization problem? Why not just change:
```
anyOf(containsString("timed out"), containsString("Das Netzwerk ist nicht erreichbar"))
```
to
```
anyOf(containsString("timed out"), containsString("Das Netzwerk ist nicht erreichbar"), containsString("No route to host (Host unreachable)"))
```

PR #786, opened 2018-03-19
test(datahandler): add assertion for couchdb test (lib-datahandler)

maxhbr:
This reverts commit 3b90fe7e6dcc56f211cb3f633e275352b988e0f6 and fixes the internationalization bug.

related to #784

### DIFF
--- a/libraries/lib-datahandler/src/test/java/org/eclipse/sw360/datahandler/couchdb/AttachmentContentDownloaderTest.java
+++ b/libraries/lib-datahandler/src/test/java/org/eclipse/sw360/datahandler/couchdb/AttachmentContentDownloaderTest.java
@@ -21,6 +21,8 @@ import org.mockito.runners.MockitoJUnitRunner;
 
 import java.io.InputStream;
 import java.io.InputStreamReader;
+import java.net.SocketException;
+import java.net.SocketTimeoutException;
 import java.util.concurrent.*;
 
 import static org.eclipse.sw360.datahandler.TestUtils.*;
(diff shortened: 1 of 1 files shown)


maierthomas:
I think we should remove this assertion based on the operating system error messages.
These error messages are language specific and currently we only verify/support German and English strings.
You may cant build the sw360portal if you have an OS with other language configuration.

maxhbr:
Sure, that is a problem, one could try to get the not-localized message or add other assertions to the type of the exception.

libraries/lib-datahandler/src/test/java/org/eclipse/sw360/datahandler/couchdb/AttachmentContentDownloaderTest.java
```diff
@@ -88,6 +87,7 @@ public String call() throws Exception {
                 } catch (ExecutionException e) {
                     Throwable futureException = e.getCause();
                     assertThat(futureException, is(notNullValue()));
+                    assertTrue(futureException instanceof SocketTimeoutException);
```
maierthomas: Maybe add `SocketException` or directly use `IOException`

maxhbr: I want to avoid `IOExcepioin`, since it contains less information.

`SocketException` occurs on your system?

maierthomas: "No route to host (Host unreachable)" should be NoRouteToHostException which extends the SocketException class

maxhbr: Oh, now I understand. It depends on the routing of the network interface on the host, which exception gets thrown.

Issue #330, opened 2017-02-01
LicenseInfo Service: template files from /etc/sw360


mcjaeger:
Redeploying the license info service losses custom license info generation templates and the ones that were available at compilation / packaging time of the *.war file is in again.

The templates shall be also considered from the /etc/sw360 folder, so the redeployment of the service does not require adaptations in /opt/sw360/webapp/licenseinfo/WEB-INF/classes .... or whatever the actual path is.

mcjaeger:
looks weirdo:
https://docs.spring.io/spring/docs/2.0.x/javadoc-api/org/springframework/ui/velocity/VelocityEngineUtils.html

maximilianhuber:
This should already work, if the templates are loaded with our util function `CommonUtils.loadResource()`. Please retest.

alex-evo:
The templates are _not_ loaded with `CommonUtils.loadResource()`. See e.g. https://github.com/sw360/sw360portal/blob/master/backend/src/src-licenseinfo/src/main/java/org/eclipse/sw360/licenseinfo/outputGenerators/LicenseInfoGenerator.java#L55

PR #790, opened 2018-03-20
feat(licenseinfo): allows velocity to load templates from etc/sw360

maierthomas:
changed velocity context configuration to load from file and class
1. **file** (/etc/sw360/)
2. **class** (./WEB-INF/classes/)

closes sw360/sw360portal#330

_edit, +template path configuration for docx generation_

### DIFF
--- a/backend/src/src-licenseinfo/src/main/java/org/eclipse/sw360/licenseinfo/outputGenerators/DocxGenerator.java
+++ b/backend/src/src-licenseinfo/src/main/java/org/eclipse/sw360/licenseinfo/outputGenerators/DocxGenerator.java
@@ -17,6 +17,7 @@ import org.apache.poi.xwpf.usermodel.XWPFParagraph;
 import org.apache.poi.xwpf.usermodel.XWPFRun;
 import org.apache.thrift.TException;
 import org.apache.xmlbeans.XmlException;
+import org.eclipse.sw360.datahandler.common.CommonUtils;
 import org.eclipse.sw360.datahandler.thrift.SW360Exception;
 import org.eclipse.sw360.datahandler.thrift.ThriftClients;
 import org.eclipse.sw360.datahandler.thrift.licenseinfo.*;
@@ -24,6 +25,7 @@ import org.eclipse.sw360.datahandler.thrift.licenses.License;
 import org.eclipse.sw360.datahandler.thrift.licenses.LicenseService;
 import org.eclipse.sw360.datahandler.thrift.licenses.Todo;
 
+import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.util.*;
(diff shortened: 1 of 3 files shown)


backend/src/src-licenseinfo/src/main/java/org/eclipse/sw360/licenseinfo/outputGenerators/DocxGenerator.java
```diff
@@ -49,7 +51,8 @@ public DocxGenerator(OutputFormatVariant outputFormatVariant, String description
     public byte[] generateOutputFile(Collection<LicenseInfoParsingResult> projectLicenseInfoResults, String projectName, String projectVersion, String licenseInfoHeaderText) throws SW360Exception {
         ByteArrayOutputStream docxOutputStream = new ByteArrayOutputStream();
         try {
-            XWPFDocument xwpfDocument = new XWPFDocument(this.getClass().getResourceAsStream(DOCX_TEMPLATE_FILE));
+            byte[] docxTemplateFile = CommonUtils.loadResource(DocxGenerator.class, DOCX_TEMPLATE_FILE).orElse(null);
```
alexbrdn: did you test what happens if `loadResource()` doesn't find the template? It looks like `orElse(null)` will cause an NPE at the next line. Maybe passing an empty byte array would be better. It would produce an empty output, but at least not fail outright. Logging a warning that the template hasn't been found would be beneficial, I imagine.

backend/src/src-licenseinfo/src/main/java/org/eclipse/sw360/licenseinfo/outputGenerators/OutputGenerator.java
```diff
@@ -93,8 +95,10 @@ public String getComponentLongName(LicenseInfoParsingResult li) {
 
     public VelocityContext getConfiguredVelocityContext() {
         Properties p = new Properties();
-        p.setProperty("resource.loader", "class");
-        p.setProperty("class.resource.loader.class", "org.apache.velocity.runtime.resource.loader.ClasspathResourceLoader");
+        p.setProperty(RuntimeConstants.RESOURCE_LOADER, "file, class");
+        p.setProperty(RuntimeConstants.FILE_RESOURCE_LOADER_PATH, "/etc/sw360");
```
alexbrdn: `"/etc/sw360"` is surely already defined as a constant. Like at `CommonUtils.SYSTEM_CONFIGURATION_PATH`, for example. You only have to make it public.

alexbrdn: Oh, and while you're at it, make it `final`, please.

backend/src/src-licenseinfo/src/main/java/org/eclipse/sw360/licenseinfo/outputGenerators/DocxGenerator.java
```diff
@@ -48,28 +50,33 @@ public DocxGenerator(OutputFormatVariant outputFormatVariant, String description
...
-            docxOutputStream.close();
-        } catch (XmlException e) {
-            throw new SW360Exception("Got XmlException while generating docx document: " + e.getMessage());
-        } catch (IOException e) {
-            throw new SW360Exception("Got IOException when generating docx document: " + e.getMessage());
-        } catch (TException e) {
-            throw new SW360Exception("Error reading sw360 licenses: " + e.getMessage());
+            return docxOutputStream.toByteArray();
+        } else {
+            throw new SW360Exception("Could not generate xwpf document for template: " + DOCX_TEMPLATE_FILE);
```
alexbrdn: this is a minor point, but I think maybe a more telling error message would be `"Could not load the template for xwpf document: " + DOCX_TEMPLATE_FILE`

backend/src/src-licenseinfo/src/main/java/org/eclipse/sw360/licenseinfo/outputGenerators/OutputGenerator.java
maierthomas: changed.
I set the variable SYSTEM_CONFIGURATION_PATH to public and final.

backend/src/src-licenseinfo/src/main/java/org/eclipse/sw360/licenseinfo/outputGenerators/DocxGenerator.java
maierthomas: corrected.

backend/src/src-licenseinfo/src/main/java/org/eclipse/sw360/licenseinfo/outputGenerators/DocxGenerator.java
maierthomas: corrected.

adrian-evo:
Tested and works fine.

Issue #773, opened 2018-03-07
For unmodified OSS, ECC classification can be done automatically:


mcjaeger:
If 3rd-party open source software sources, in particular cryptographic algorithms therein, were not modified then there is no need for a manual ECC classification if that OSS is publicly available. For this case the AL and ECCN values can be set to value 'N' automatically.

Thus the following procedure shall be added in sw360:

If there is a download URL (from an external source) given (=implying that OSS was taken from there)  and the ECC fields are empty, then
* the ECCN classification shall be 'N' automatically
* the AL classification shall be 'N' automatically and
* the 'ECC approved' field shall be checked automatically.

PR #798, opened 2018-03-26
feat(release): automatically update ecc information if download url is set

maierthomas:
tbd

closes sw360/sw360portal#773


mcjaeger:
I think the three fields shall be from the user who actually enters the information (either at creation or update of the attribute release date)

maierthomas:
Current implementation:
- "automatically set, public OSS"
- "no department"
- "date of the chages"

In the steps `addRelease` and `updateRelease` we have the -email address of the user.
We can consider to use this address instead of the constant text for contact person.

mcjaeger:
@maierthomas actually I see I was wrong. Actually department is the old name. I think it should be "no group" but the assessor contact person is not conflicting since it is supposed to be an e-mail in this fields?

maierthomas:
contact person field is only a string value of mail address

- email of user
- "no group"
- date of change

adrian-evo:
When the download url is changes through moderation, the fields are not updated. The moderation request list the fields to be updated, excluding Contact Person, Department, Date, and after approving the moderation request, the fields remains unchanged.

mcjaeger:
two cases:

1. user can write to release document because of creator role or clearing admin role (or higher): download URL is written and the sw360 application writes the three ECC attributes accordingly.

2. user cannot write but creates a moderation request. then approving the moderation request writes the download URL and the sw360 application sets also the ECC values accordingly. No extra moderation request is sent to ECC admin.

Please keep in mind that auto - setting of download URL is for saving work for ECC admins.

maierthomas:
Exactly this behavior is implemented.

- download URL for all steps set, eg. http://sw360.org
1. User has made **changes in ECC** fields / User has **ECC Write access** --> sw360 sets **no ECC values** (because user changed the ECC values)
2. User has made **no changes in ECC** fields (fields are empty or default) / User **has Write access** --> sw360 sets **automatically the ECC values** 
3. User has made **no changes in ECC** fields (fields are empty or default) / User has **no Write access** --> sw360 sets **automatically the ECC values** in the moderation request (**no ECC moderation request**)
3. User has made **changes in ECC** fields (fields are empty or default) / User has **no Write access** --> sw360 sets **no ECC values** in the moderation request (**ECC moderation request**)

maierthomas:
+ fixed escaping problem in moderation list for closed moderations entries

adrian-evo:
Both cases works as described above. However, the moderation request is sent not only to admin, but also to creator of the release. E.g. if user1 creates the release, then user2 updates the download URL, user1 also receives this moderation request which seems confusing.

mcjaeger:
@adrian-evo I am confused. my idea was, it is normal that the creator of a document gets also the modertation request. (in additon to moderators and clearing admins and admins)

backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ComponentDatabaseHandler.java
```diff
@@ -649,11 +659,34 @@ public boolean hasChangesInEccFields(Release release, Release actual) {
                 .reduce(false, Boolean::logicalOr);
     }
 
-    private void autosetEccUpdaterInfo(Release release, User user) {
+    public boolean hasEmptyEccFields(Release release) {
+        EccInformation eccInformation = release.getEccInformation();
+        return (eccInformation.getAL() == null || eccInformation.getAL().isEmpty()) &&
```
alexbrdn: `isNullOrEmptyWhitespace()`, please

backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ComponentDatabaseHandler.java
```diff
@@ -649,11 +659,34 @@ public boolean hasChangesInEccFields(Release release, Release actual) {
...
+    }
+
+    private void autosetEccFieldsForReleaseWithDownloadUrl(Release release, String contactPerson) {
+        // For unmodified OSS, ECC classification can be done automatically: sw360/sw360portal#773
+        String url = release.getDownloadurl();
+        if (url != null && !url.isEmpty()) {
+            if (CommonUtils.isValidUrl(url)) {
+                autosetEccUpdaterInfo(release, contactPerson, ASSESSOR_DEFAULT_GROUP);
+                EccInformation eccInfo = release.getEccInformation();
+                eccInfo.setAL("N");
```
alexbrdn: magic constant

backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ComponentDatabaseHandler.java
```diff
@@ -649,11 +659,34 @@ public boolean hasChangesInEccFields(Release release, Release actual) {
...
+                eccInfo.setECCN("N");
```
alexbrdn: magic constant

frontend/sw360-portlet/src/main/webapp/html/moderation/view.jsp
```diff
@@ -171,7 +171,7 @@
                 "5": '<sw360:DisplayUserEmailCollection value="${moderation.moderators}" bare="true"/>',
                 "6": "<sw360:DisplayEnum value="${moderation.moderationState}"/>",
                 <core_rt:if test="${isUserAtLeastClearingAdmin == 'Yes'}">
-                "7": "<img class='delete' src='<%=request.getContextPath()%>/images/Trash.png' onclick=\"deleteModerationRequest('${moderation.id}','<b>${moderation.documentName}</b>')\"  alt='Delete' title='Delete'>"
+                "7": "<img class='delete' src='<%=request.getContextPath()%>/images/Trash.png' onclick=\"deleteModerationRequest('<sw360:out value="${moderation.id}"/>','<b><sw360:out value="${moderation.documentName}"/></b>')\"  alt='Delete' title='Delete'>"
```
alexbrdn: why use `<sw360:out>`  on moderation.id?

backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ComponentDatabaseHandler.java
maierthomas: corrected

backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ComponentDatabaseHandler.java
maierthomas: corrected

backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ComponentDatabaseHandler.java
maierthomas: corrected

Issue #799, opened 2018-03-29
sw360 project build fails in the first time cause cannot build authorization server


maierthomas:
sw360 project build fails in the first time cause cannot build authorization server
please verify pom settings for authorization server.

**test env**: vagrant setup (initial project build)

default: [ERROR] Failed to execute goal org.springframework.boot:spring-boot-maven-plugin:1.5.8.RELEASE:repackage (default) on project authorization-server: Execution default of goal org.springframework.boot:spring-boot-maven-plugin:1.5.8.RELEASE:repackage failed: Source must refer to an existing file, got /opt/apache-tomcat-7.0.67/deploy/authorization.war -> [Help 1]
    default: [ERROR] 
    default: [ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
    default: [ERROR] Re-run Maven using the -X switch to enable full debug logging.
    default: [ERROR] 
    default: [ERROR] For more information about the errors and possible solutions, please read the following articles:
    default: [ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/PluginExecutionException
    default: [ERROR] 
    default: [ERROR] After correcting the problems, you can resume the build with the command
    default: [ERROR]   mvn <goals> -rf :authorization-server

PR #802, opened 2018-03-29
fix(rest): seperate creation of rest fatjars to maven profile

maierthomas:
- removed the auto build for fatjars in rest projects
--> added maven profile to obtain the option to build rest fatjars

closes sw360/sw360portal#799

### DIFF
--- a/rest/authorization-server/pom.xml
+++ b/rest/authorization-server/pom.xml
@@ -129,24 +129,6 @@
         </dependencies>
     </dependencyManagement>
 
-    <build>
-        <finalName>authorization</finalName>
-        <plugins>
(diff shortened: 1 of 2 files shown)


mcjaeger:
maybe testing could be like this using the fatjar:
`java -jar myfatjar.jar`

maierthomas:
Yes you can run the fatjar (wars) successfully with `java -jar resource.war/authorization.war`
In general if someone needs this fatjar war files the user can generate this fatjar with maven profile parameter fatjar.
`maven package -P fatjar`
I think we should describe this options in the readme or wiki page.

Issue #794, opened 2018-03-22
Missing properties in REST API


tngraf:
Some important properties are not yet available via REST API:
* Project.Tag
* Project.PreevaluationDeadline
* Project.SystemTestStart
* Project.SystemTestEnd
* Component.Categories
* Component.Languages
* Component.OperatingSystems

PR #803, opened 2018-04-03
dev(rest): add missing rest properties for components and projects

maierthomas:
- add properties for rest API and enhanced documentation

closes sw360/sw360portal#794


adrian-evo:
Tested, the mentioned properties were fetched successfully.

Issue #805, opened 2018-04-03
Adding usage count of releases in addition to searchByReleaseId  in Projects.thrift


mcjaeger:
When using the method to get some release usage info there is in
https://github.com/sw360/sw360portal/blob/master/libraries/lib-datahandler/src/main/thrift/projects.thrift

```
...
    /**
     * list of projects which are visible to the `user` and match the `name`
     */
    list<Project> searchByName(1: string name, 2: User user);

    /**
     * list of short project summaries which are visible to the `user` and have `id` in releaseIdToUsage
     */
    set<Project> searchByReleaseId(1: string id, 2: User user);
```
but this does not provide the big picture since it is limited to the user's project visibility limitations. For this another method is required that provides the total count of projects using the release (or releases) regardless of project access rights.

Note that there are two methods for a single release id and an array of release ids, the latter one is used to query for all releases of a component at a single time.

At the same time, the issue with linkedProjectSummaryAccordingToAccessibility (dropping the second add) of the project list should be corrected.

PR #808, opened 2018-04-04
dev(components): hide inaccessible projects in component summary

maierthomas:
- hide inaccessible projects in the "is used by the following" section for a component
- add usage count of linked projects (`allUsingProjectsCount`)

closes sw360/sw360portal#805

### DIFF
--- a/backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ProjectRepository.java
+++ b/backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ProjectRepository.java
@@ -1,5 +1,5 @@
 /*
- * Copyright Siemens AG, 2013-2016. Part of the SW360 Portal Project.
+ * Copyright Siemens AG, 2013-2018. Part of the SW360 Portal Project.
  *
  * SPDX-License-Identifier: EPL-1.0
  *
@@ -10,7 +10,6 @@
  */
 package org.eclipse.sw360.datahandler.db;
 
-import com.google.common.collect.Sets;
 import org.eclipse.sw360.components.summary.ProjectSummary;
 import org.eclipse.sw360.components.summary.SummaryType;
 import org.eclipse.sw360.datahandler.couchdb.DatabaseConnector;
@@ -21,9 +20,7 @@ import org.eclipse.sw360.datahandler.thrift.users.User;
 import org.ektorp.support.View;
 import org.jetbrains.annotations.NotNull;
 
-import java.util.HashSet;
-import java.util.List;
-import java.util.Set;
+import java.util.*;
 import java.util.stream.Collectors;
 
 import static com.google.common.base.Strings.isNullOrEmpty;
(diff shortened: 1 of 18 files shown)


frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/portlets/components/ComponentPortlet.java
```diff
@@ -760,19 +761,21 @@ private void prepareDetailView(RenderRequest request, RenderResponse response) {
     private void setUsingDocs(RenderRequest request, User user, ComponentService.Iface client, Set<String> releaseIds) {
         Set<Project> usingProjects = null;
         Set<Component> usingComponentsForComponent = null;
+        int allUsingProjectsCount = -1;
```
maxhbr: Why is `-1` the fallback, instead of `0`?

maxhbr: instead of using `-1` as fallback one could only fill it after line 769, i.e. if the call was successful

maxhbr: Oh, now I see https://github.com/sw360/sw360portal/pull/808/files#diff-e8bd4c4dc1849fc1be46b0781739dcd5R13. I think that this is not the nicest way of handling exceptions.

backend/src-common/src/main/java/org/eclipse/sw360/datahandler/db/ProjectRepository.java
```diff
@@ -153,17 +155,15 @@ public ProjectRepository(DatabaseConnector db) {
         return makeSummaryFromFullDocs(SummaryType.SHORT, projectsMatchingNameAndVersion);
     }
 
-
     @View(name = "byreleaseid", map = BY_RELEASE_ID_VIEW)
     public Set<Project> searchByReleaseId(String id, User user) {
-        Set<String> searchIds = queryForIdsAsValue("byreleaseid", id);
-
-        return new HashSet<>((makeSummaryFromFullDocs(SummaryType.SHORT, filterAccessibleProjectsByIds(user, searchIds))));
+        Set<String> ids = new HashSet<>(Arrays.asList(id));
```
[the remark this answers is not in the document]
maierthomas: corrected

frontend/sw360-portlet/src/main/java/org/eclipse/sw360/portal/portlets/components/ComponentPortlet.java
maierthomas: `usingProjectTable` evalutes the value of `allUsingProjectsCount`.
You are right, its not the best way of handling exceptions. 
`setUsingDocs` should throw the exception instead of only logging it. 
But I dont want to refactor this method here.

mcjaeger:

maierthomas:
@mcjaeger updated branch again,
can you please redeploy on stage for test purpose?

maierthomas:
tested successfully on local and stage instance

Issue #329, opened 2017-01-30
Validate license id with short name


toedter:
Currently the short name of a license is used as (couchbase) id. But the API does not validate if id and short name match when a new license is created. So it is possible to create a license with id "apache20" and shortname "mit".

mcjaeger:
(from discussion) short name should be deprecated and not used anymore.

mcjaeger:
apparently will be solved when we will have a commit from an undisclosed source using SPDX tools and SPDX license list import

maximilianhuber:
> apparently will be solved when we will have a commit from an undisclosed source using SPDX tools and SPDX license list import

No, this is not the topic of #370.

To solve the issue mentioned by @toedter one has to check whether the ID equals the shortname (if set) everytime one creates or updates a license.

An alternative more complete solution would be to make the ID of the License an UUID again. But this will then interfere with https://github.com/sw360/sw360portal/issues/19 which was implemented via https://github.com/sw360/sw360portal/pull/135.

PR #809, opened 2018-04-05
fix(license): validate license id and short name

maierthomas:
- add validate function to ThriftValidate.java
- verify license id / short name in addLicense and updateLicense

closes sw360/sw360portal#329

### DIFF
--- a/libraries/lib-datahandler/src/main/java/org/eclipse/sw360/datahandler/thrift/ThriftValidate.java
+++ b/libraries/lib-datahandler/src/main/java/org/eclipse/sw360/datahandler/thrift/ThriftValidate.java
@@ -1,5 +1,5 @@
 /*
- * Copyright Siemens AG, 2014-2017. Part of the SW360 Portal Project.
+ * Copyright Siemens AG, 2014-2018. Part of the SW360 Portal Project.
  *
  * SPDX-License-Identifier: EPL-1.0
  *
@@ -25,6 +25,7 @@ import org.eclipse.sw360.datahandler.thrift.vendors.Vendor;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
+import java.util.regex.Pattern;
 
 import static com.google.common.base.Predicates.notNull;
 import static org.eclipse.sw360.datahandler.common.SW360Assert.*;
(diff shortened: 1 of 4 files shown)


maierthomas:
`assertId` method in `SW360Asserts.java` should validate the correct document id?
How do I validate the doc id exactly?

mcjaeger:
please consider `1*(ALPHA / DIGIT / "-" / "." )`

from https://spdx.org/spdx-specification-21-web-version#h.jxpfx0ykyb60

ID string is like "Apache-2.0" or "MIT" or "GPL-2.0-only"

mcjaeger:
(from discussion, Alex proposed) ... to use a new "assertLicenseId" not just assertId

maierthomas:
Should we include also `"+"` or is this character deprecated?
The validation of the value is only for a **new** license

maierthomas:
(from discussion) the character `"+"` is allowed

adrian-evo:
Tested and the license short name is validated fine.

libraries/lib-datahandler/src/main/java/org/eclipse/sw360/datahandler/thrift/ThriftValidate.java
```diff
@@ -244,6 +245,28 @@ public static void validateAttachment(AttachmentContent attachment) throws SW360
         attachment.setType(TYPE_ATTACHMENT);
     }
 
+    public static void validateLicenseIdMatch(License license) throws SW360Exception {
+        if (license.isSetId() && license.isSetShortname()) {
+            String message = "license short name must be equal with license id";
```
alexbrdn: I believe "equal to" is the grammatically correct phrase

libraries/lib-datahandler/src/main/java/org/eclipse/sw360/datahandler/thrift/ThriftValidate.java
```diff
@@ -244,6 +245,28 @@ public static void validateAttachment(AttachmentContent attachment) throws SW360
...
+            assertEquals(license.getId(), license.getShortname(), message);
+        }
+    }
+
+    public static void validateLicense(License license) throws SW360Exception {
+        final Pattern pattern = Pattern.compile("[A-Za-z0-9\\-.+]*");
```
alexbrdn: please extract the regexp into a named constant

alexbrdn: maybe even the whole pattern

libraries/lib-datahandler/src/main/java/org/eclipse/sw360/datahandler/thrift/ThriftValidate.java
maierthomas: corrected

maierthomas:
- renamed `validateLicense` to `validateNewLicense`
- removed if condition (isSet) for `license.shortname`
- removed assertId and pattern check for `license.id`

PR #811, opened 2018-04-09
fix(moderation): set all using project count to merge pages

maierthomas:
fix the error on moderation page

closes sw360/sw360portal#111


mcjaeger:
first deployment seems to crash with 
`java.lang.IllegalStateException: java.io.FileNotFoundException: /opt/sw360/temp/0-sw360-portlet/WEB-INF/lib/datatables.net-bs-2.1.1.jar (No such file or directory)` ... is this a deployment issue?

mcjaeger:
tested by Mr. Jaeger on clone of productive ... works

PR #812, opened 2018-04-09
fix(release): remove contact person and department and add ecc comment

maierthomas:
- remove the autoset values for assessor contact person and assessor department
- add ecc comment: automatically set because download URL indicates original OSS


maierthomas:
@mcjaeger please retest.
autoset of ECC values is working for both scenarios (moderation request, full permission).
Also on both cases the contact person and department is unset.

adrian-evo:
Tested and assessor data is not set anymore. However:
1.  ECC comment is only "automatically set" and not "automatically set because...". Is this then intention?
2. Moderation request shows only that url will change in Proposed changes. Previously also ECC data was listed as a change. Is this ok?

mcjaeger:
@adrian-evo thanks for your feedback, to your questions:
1. yes, it was not displaying properly otherwise
2. actually the "N" in ECC data should be there, but not moderator e-mail and group (which was written before but was not good, because the system was setting it not the person merging the moderation request)

PR #815, opened 2018-04-13
fix(docker): Change thrift url to working link

maxhbr:
The old URL only returns 404


alexbrdn:
Unfortunately, I cannot test this at the moment, but considering how trivial the change is, I suggest that you merge it directly, provided that you have tested it once in your environment.

maxhbr:
I have also tested this PR

PR #816, opened 2018-04-13
fix(ui): ui should be sticky

maierthomas:
moving to eclipse repo -> closed


maierthomas:
moving to eclipse repo -> closed

PR #818, opened 2018-04-19
feature: codescoop integration autocomplete (demo)

sweetca:
Just demo of UI integration of CodeScoop service


sweetca:
@mcjaeger this is UI integration example for such features as:
- Autosearch & Autofill when adding new components
- Component metadata from codescoop on component list page

